PR #1, opened 2026-03-16
Add wildcard-match-arm lint

martijnarts:
First Whisker lint: `wildcard_match_arm` flags wildcard patterns in match arms when the scrutinee is an enum. Wildcards are allowed on non-enum types (integers, bools) and on `#[non_exhaustive]` enums from external crates, but flagged on local `#[non_exhaustive]` enums since the author controls all variants.

This also introduces `whisker_testing`, a test harness that wraps `compiletest_rs` and the dylint driver to run both stderr snapshot diffing (UI mode) and inline `//~` annotation checking (compile-fail mode with `-D warnings`), matching how Clippy tests its own lints.

All six Tracey spec rules have full coverage: implemented and verified.


martijnarts:
Superseded by #17 and #21

PR #37, opened 2026-04-10
Restructure specs for tree-sitter domain design

martijnarts:
Breaks the monolithic lints.md specification into per-lint spec files and introduces new domain specifications for the core engine, macro system, language SDK, and Rust SDK.

The architecture document describes the new layered design: source code flows through tree-sitter parsing, decoration providers attach semantic information, and rules produce diagnostics from the decorated syntax tree.

The Tracey configuration is updated to map each new spec to its corresponding implementation crate.

Part of the domain design series. This PR is independent and can be reviewed/merged in any order.


martijnarts:
We're not using Tracey in the future.

PR #197, opened 2026-08-03
Document the platform architecture

martijnarts:
Whisker has settled into its tree-sitter shape — a language-agnostic pipeline that parses, decorates, and walks syntax trees, with rust-analyzer supplying semantic decorations for Rust — but that design only lived in the code. This adds an ARCHITECTURE.md that writes it down.

The document leads with the split that motivates everything else: syntax is cheap, so tree-sitter parses every file in isolation; semantics is expensive, so it enters only as decorations a provider computes up front and attaches to nodes. From there it walks the parse-decorate-execute-report pipeline, gives each crate its role and its place in the dependency order, and explains the pieces that aren't obvious from a first read — why the RustLintPass trait is generated from the grammar's node-types.json at build time, how supertype dispatch works, and why the provider collects targets in one pass and resolves them in another (tree-sitter and rust-analyzer keep separate trees, and byte offsets are the only thing they agree on).

It closes with the extension points for a new lint, a new decoration, and a new language. There is one diagram, for the pipeline; everything else reads better as prose.

https://claude.ai/code/session_01FEuqJxbEqJa5jh5qnuh6fj

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,0 +1,293 @@
+# Architecture
+
+Whisker is a language-agnostic linting platform built on [tree-sitter][ts].
+It ships with Rust lints that enforce Aonyx's coding conventions — rules
+that Clippy doesn't cover, like derive ordering and wildcard match arms.
+
+This document describes how the platform works as a whole. It does not
+describe individual lints; see `specs/lints.md` for those.
+
+## The core idea
+
+Most of what a lint needs to know is syntactic, and syntax is cheap: a
+tree-sitter grammar parses any source file in isolation, without a
+toolchain, a build, or even a valid project. Some lints, however, need
+semantic facts a syntax tree cannot provide — "is this match scrutinee an
+enum?", "does this function return `Result`?" — and semantics is expensive,
+requiring a real language toolchain with knowledge of the whole project.
+
+Whisker separates the two. The platform parses files with tree-sitter and
+walks the syntax tree through lint rules. Semantic information enters the
+picture only as **decorations**: annotations that a language-specific
+provider computes up front (for Rust, using [rust-analyzer][ra] as a
+library) and attaches to individual syntax nodes. Lint rules read
+decorations off the tree; they never talk to a toolchain directly.
+
+This split keeps rules simple and fast to test — a rule is a pure function
+of a decorated tree, so tests can hand-construct decorations instead of
+spinning up rust-analyzer — while still letting rules make type-aware
+decisions.
+
+## The pipeline
+
+Every file flows through four stages, orchestrated by
+`whisker_core::Pipeline`:
+
+```mermaid
+flowchart LR
+    src[source file] --> parse[parse<br>tree-sitter]
+    parse --> decorate[decorate<br>providers]
+    ra[rust-analyzer] --> decorate
+    decorate --> execute[execute<br>tree walker × lint passes]
+    execute --> render[report<br>codespan-reporting]
+```
+
+1. **Parse.** The file is parsed with the tree-sitter grammar for its
+   language (detected from the file extension). The result is wrapped in a
+   `DecoratedTree`: the tree-sitter tree plus the source text, the file
+   path, and an initially empty `DecorationMap`.
+2. **Decorate.** Each registered `DecorationProvider` inspects the tree and
+   inserts decorations into the map, keyed by tree-sitter node ID. This is
+   the only stage that touches a language toolchain.
+3. **Execute.** The tree walker performs a depth-first traversal of all
+   named nodes. At every node it calls `check_node` on each `LintPass`,
+   collecting the diagnostics they return. Passes take `&mut self`, so a
+   rule may carry state; the CLI constructs a fresh set of passes per file
+   so no state leaks across files.
+4. **Report.** Diagnostics are rendered to stderr with
+   [codespan-reporting][csr]: the primary span becomes the primary label,
+   origin and related locations become secondary labels, and suggestions
+   become notes.
+
+## Workspace layout
+
+The platform lives in `crates/`; each lint rule is its own crate in
+`lints/`. The root `Cargo.toml` lists every lint crate as an explicit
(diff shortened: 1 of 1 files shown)


ARCHITECTURE.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Development workflow
+
+The development environment is managed with [Flox][flox], and the
+`justfile` runs every recipe inside it. `just pre-commit` runs the full
+gauntlet: formatting (rustfmt with unstable options, prettier, taplo),
+linting (clippy with `-D warnings`, markdownlint, zizmor), tests, and
+Tracey validation. Tests lean on property-based testing
+with proptest, and every custom type carries `Send`/`Sync`/`Unpin` trait
+tests by convention.
```
martijnarts: This is README stuff or CONTRIBUTORS maybe, or AGENTS.md. Anyway, not architecture.

ARCHITECTURE.md
```diff
@@ -0,0 +1,293 @@
...
+`create_lint_passes`.
+
+**A new decoration** is a type (derived with `#[derive(Decoration)]` to
+fix its cardinality), a new `Target` variant in the provider's collection
+walk, and resolution logic against the semantic model.
+
+**A new language** is a `Language` variant, a tree-sitter grammar with a
+build-script invocation of `whisker-codegen` to generate its
+`{Language}LintPass` trait, an adapter, and — if its lints need
+semantics — a `DecorationProvider` backed by that language's toolchain.
```
martijnarts: Also not architecture

ARCHITECTURE.md
```diff
@@ -0,0 +1,293 @@
...
+`whisker check [path]`:
+
+1. Walks the target directory (or takes a single file), collecting every
+   file whose extension maps to a supported language.
+2. Loads the `RustDecorationProvider` once for the target workspace.
+3. Runs the pipeline on each file with a fresh set of lint passes.
+4. Renders all collected diagnostics and exits non-zero if any diagnostic
+   has severity `Error` or above, or if any file failed under
+   `--keep-going` (without the flag, the first failure aborts the run).
+
```
martijnarts: too specific for an architecture doc

ARCHITECTURE.md
```diff
@@ -0,0 +1,293 @@
...
+passes through the real tree walker. Because decorations are plain values,
+a test can assert a semantic lint's behavior — "this scrutinee is an
+external non-exhaustive enum" — without a Cargo project or rust-analyzer
+in sight. `assert_diagnostic` provides a fluent builder for checking rule
+ID, severity, message, span, and attachment counts, and `fixtures` loads
+directory-based test cases.
+
+The provider itself is tested separately: `whisker-rust` has an
+integration test that loads a real fixture Cargo project once and checks
+the decorations rust-analyzer produces for it.
```
martijnarts: ALL of these crate-specific descriptions is way too specific for architecture doc. This should be crate level documentation if anything.

ARCHITECTURE.md
```diff
@@ -0,0 +1,293 @@
...
+## The pipeline
+
+Every file flows through four stages, orchestrated by
+`whisker_core::Pipeline`:
+
+```mermaid
+flowchart LR
+    src[source file] --> parse[parse<br>tree-sitter]
+    parse --> decorate[decorate<br>providers]
+    ra[rust-analyzer] --> decorate
```
martijnarts: should probably be "language provider" not "rust-analyzer" to make it agnostic.

martijnarts (commented):
overall this is pretty good, but we can be more concise and careful with our wording, less implementation specific and more architecture higher level overview.

ARCHITECTURE.md
martijnarts: Done — the diagram now shows a generic `language provider` feeding the decorate stage. rust-analyzer is named once in the body, as the instance behind Rust's provider rather than as part of the pipeline itself.

ARCHITECTURE.md
martijnarts: Agreed, all seven crate subsections are gone. What was genuinely architectural in them — the generated lint pass trait, the provider seam, decoration cardinality — is now stated as a concept under "Language support" instead of as a tour of the crates that implement it. The crate table survives as the map.

ARCHITECTURE.md
martijnarts: Cut. The CLI's file-walking and exit-code sequence was implementation detail; what's left is that rules are linked in explicitly, which is the part that actually shapes the system.

ARCHITECTURE.md
martijnarts: Cut. Worth noting CLAUDE.md already documents this — "Development environment" and the "Quick reference" recipes — so it was duplication as well as misfiled.

ARCHITECTURE.md
martijnarts: Cut as well. If it's wanted, a CONTRIBUTING.md is the better home; I'd rather not invent one in this PR.

martijnarts:
Rewrote it along the lines you asked for — 293 lines down to 115, in fbac8d1.

The organizing principle I used for the cut: keep decisions, drop the tour. What survives is the syntax/semantics split that motivates the design, the pipeline, the two seams a language plugs into, how rules behave, and the crate map. What went is everything that was really describing an implementation — the per-crate walkthroughs, the CLI's checking sequence, the extension recipes — plus the development workflow, which CLAUDE.md already covers.

The one judgment call worth flagging: I kept the generated-lint-pass-trait idea and the decoration cardinality rule, but restated them as concepts under "Language support" rather than as descriptions of `whisker-codegen` and `whisker-types`. Those felt structural rather than incidental — the trait being generated from the grammar is why rules can't drift from it — but they were living inside the crate sections you flagged, so say the word if you'd rather they went too.

I also dropped the "Specs and traceability" section here rather than in the stacked PR. It's process rather than architecture, so it fell under the same principle as the development workflow.

#198 is rebased on top of this and no longer touches ARCHITECTURE.md at all — the rewrite already removed everything it needed to remove. Both branches are green.

PR #198, opened 2026-08-03
Remove Tracey spec coverage tracking

martijnarts:
Stacked on #197 — review that one first.

The specs Tracey tracked had stopped describing whisker. They were written for the Dylint-era rustc driver and its compiletest harness, both of which the tree-sitter rewrite replaced, so `specs/whisker.md` still documented a `RUSTC_WORKSPACE_WRAPPER` and a lint store that no longer exist, and the r[impl] annotations scattered through the code pointed at requirements that had quietly stopped matching it. Traceability to a stale document is worse than none, because it reads as authoritative.

Rather than rewrite the specs to chase the implementation, this removes them along with Tracey itself: the config, the two justfile recipes and their pre-commit hookup, the .gitignore entry, and all 79 r[impl] and r[verify] annotations. Each rule already documents its behavior in its doc comment and pins it with tests, which is where the truth actually lives.

Two things came off with it. The compiletest `ui/` fixtures under three lint crates were orphaned by the same rewrite — nothing has run them since, they carry `//~ ERROR` annotations only compiletest_rs understands, and their only remaining content was r[verify] markers — so they go. And cargo-binstall entered the Flox environment in the same commit that added Tracey and exists solely to install it, so it leaves too.

One loose end: `.flox/env/manifest.lock` embeds a copy of the activation hook that is already stale on main (it still carries a dylint install line that manifest.toml dropped long ago). Regenerating it needs a working Nix daemon, which I don't have locally, so I've left it untouched rather than hand-edit a lockfile. Worth a follow-up.

Verified with cargo build, clippy -D warnings, rustfmt, and the full test suite, all green.

https://claude.ai/code/session_01FEuqJxbEqJa5jh5qnuh6fj


martijnarts:
The `lint-markdown` failure turned out to be a latent bug this branch exposed rather than a new one.

The recipe passed an unquoted `**/*.md` to markdownlint, and just runs recipes under `sh`, where `**` is just an ordinary glob. So the shell expanded it to `*/*.md` — one directory deep — and markdownlint has only ever seen `specs/`. README.md, CLAUDE.md, and CHANGELOG.md have never been linted by CI. Deleting the specs left the pattern matching nothing, the shell passed it through literally, markdownlint did its own recursive expansion, and the whole repo got linted for the first time.

Pushed 244bd8c: quoted the glob so that recursive behavior is deliberate rather than an accident of what happens to sit one level down, and gave the three trycmd fixtures it newly reaches a `console` language on their fences. I checked that trycmd still executes them rather than silently skipping — corrupted an expected version string, watched the test fail, restored it, watched it pass.

Worth knowing for the review: this means CI is now linting several Markdown files it never has before. Locally those three fences were the only violations across the repo.

PR #203, opened 2026-08-05
Add a deny-warnings flag to whisker check

martijnarts:
All six lints emit `Severity::Warn`, and `check` exited 1 only on diagnostics at
`Severity::Error` or above. `whisker check` could therefore print any number of
violations and still exit 0. That was harmless while the CLI ran no lints at all,
but now that it runs all six it means whisker cannot fail a build, which is most
of what a linter is for.

This change adds `--deny-warnings`, mirroring `cargo clippy -- -D warnings`. The
flag arrives from clap as a `bool` because that is what a command line parses,
and is converted into a `FailureThreshold` as soon as the arguments are
destructured, so nothing below that boundary has to work out what a bare boolean
means.

The more useful part is what moved. Deciding whether a run failed used to happen
inline, next to `std::process::exit(1)`, where no test could reach it. It now
lives in `CheckOutcome`, which is a pure function of the diagnostics and the
threshold, so the cases that matter are ordinary unit tests: no diagnostics,
warnings only with the flag off and on, errors with the flag off, mixed
severities, and `Help` and `Info` diagnostics, which must never fail a build
under either setting.

Per-file failures route through one type for the same reason. A single file can
fail because its bytes are not readable as UTF-8 or because the pipeline rejects
the source that was read, and both have to reach the exit code identically.
`ErrorRecovery` is the one place that decides whether such a failure ends the
walk, so the two paths cannot drift apart or disagree about whether the run
failed.

Two things are deliberately not here. The lints keep their `Warn` severity
rather than being promoted, because which rules are worth failing on is a
project's decision and `specs/whisker.md` already anticipates configuring it.
And `discover_files` still reports zero files as a successful run; that hole
belongs to the change that rewrites discovery, which stacks directly on top of
this one.


martijnarts:
Superseded by #208, which carries the same change on a renumbered branch as part of the resplit stack.

PR #204, opened 2026-08-05
Skip ignored files during whisker check

martijnarts:
File discovery walked everything. Pointed at this repository it would have
linted `target/`, plus `examples/sample_project` and
`crates/whisker-rust/tests/fixtures/sample_project`, both of which are
deliberately full of violations because they are lint fixtures. It also
swallowed every walk error through `filter_map(|e| e.ok())`, so a directory
whisker could not read simply vanished from the run.

This replaces the walk with the `ignore` crate, which honors `.gitignore`,
`.ignore`, the user's global gitignore, and hidden-file rules. A linter that
lints `target/` is broken independently of anything else, so this is worth doing
on its own terms. Layered on top is a `[workspace.metadata.whisker]` table in the
workspace manifest holding gitignore-syntax patterns, which is how this
repository excludes its own fixtures. The table is deliberately minimal —
`specs/whisker.md` anticipates configuring rule severity and enablement, and
none of that is built here — but a malformed table is an error rather than an
empty default, because a typo'd key that silently ignores nothing is the same
class of bug as everything else this change closes.

Three silent successes go with it. Discovering zero files is now an error naming
both causes, since printing nothing and exiting 0 is byte-identical to a clean
run and this change is what makes an empty result an ordinary outcome. An
explicitly named file whose extension has no grammar is rejected rather than
handed to the Rust parser and pronounced clean; `whisker check Cargo.toml` used
to exit 0. And ignore-file parse errors, which the `ignore` crate reports by
attaching them to the yielded entry rather than by yielding an error, now go
through the same policy as any other walk failure instead of being dropped.

Excluding the two fixture workspaces is load-bearing rather than cosmetic. Both
sit outside the root workspace, so rust-analyzer has no data for them, and the
change stacked directly on top of this one turns "no provider covers this file"
into a hard error. A unit test loads this repository's own manifest and asserts
no discovered file falls under them, and that whisker's own source still does, so
it cannot pass vacuously.

`crates/whisker/tests/fixtures` is deleted. Those were loose `.rs` files
belonging to no crate, which let a test assert a clean run over source no
decoration provider ever understood. The tests that used them now build real
temporary Cargo packages instead.


martijnarts:

PR #205, opened 2026-08-05
Make a decoration provider state its coverage

martijnarts:
`DecorationProvider::decorate` returned `anyhow::Result<()>`, so `Ok(())` meant
both "I decorated this file" and "I have nothing for this file". The pipeline
read the first meaning, walked a completely undecorated tree, and the three
lints that depend on semantic decorations found nothing. Whisker reported the
file clean because it had been unable to analyze it, which is indistinguishable
to the user from having found no problems.

Providers now return `Coverage`, and a provider that declines hands back a
`CoverageGap` explaining why. `Pipeline::run_on_source` refuses to walk a tree no
provider claimed. Decorations come back by value rather than being written
through a mutable handle, which makes "declined after writing half its
decorations" unrepresentable rather than merely discouraged.

Establishing that contract surfaced three further defects, each of which had the
same shape as the one above.

VFS membership was the wrong predicate. `Semantics::parse_guess_edition` papers
over a missing module by guessing the current edition and then resolving
nothing, so a `.rs` file the toolchain has interned but no crate reaches passed
the guard, parsed, and produced no decorations. Module membership is the real
question.

Cargo leaves `test` out of a crate's cfg options by default, so every
`#[cfg(test)] mod tests` block resolved to nothing while whisker walked it and
reported the file clean. Roughly half the lines of this codebase live inside such
blocks, which made this the difference between linting the project and linting
the part of it that is not tested. Editors running rust-analyzer set the flag for
exactly this reason, and now so does whisker.

A file whose bytes are not UTF-8 is interned by the loader without its text ever
being recorded, and rust-analyzer panics when something later reads it — while
building the definition map of the crate around it, or while inferring a body,
so the file that brings the process down is never the one whisker was asked
about. The load now repairs such files, and the excluded-file guard is a guard
rather than a courtesy, because an excluded file keeps its identity and loses
both its contents and its source root.

Stopping at a file that cannot be analyzed no longer discards the diagnostics
already collected from the files before it. That work was done and those findings
were true; losing them made one bad file cost the user twice.


martijnarts:

PR #206, opened 2026-08-05
Resolve error types by identity, not by rendering

martijnarts:
`anyhow_missing_context` exists to catch `foo()?` used without `.context(...)` in
a function returning `anyhow::Result`. It had never once fired on that form.

The provider resolved a try-operand's type from the operand's start offset, and
took the innermost expression at that offset. For `std::fs::read_to_string("x")?`
the token at the start is `std`, whose innermost expression ancestor is the
callee path, not the enclosing call. The resolved type was therefore a function
type, `is_result()` was false, and the lint bailed at its second guard. The same
mistake was available to match scrutinees and if-else branches, and to any
receiver, index, field access, or parenthesized expression, so resolution now
uses each node's full byte range rather than the token at its start.

Making the rule fire revealed that it had also never been correct. It decided
whether a function returned an anyhow error by comparing the rendered type name
against `"anyhow::Error"` or `"Error"`, and rust-analyzer renders `syn::Error`,
`serde_json::Error`, `std::io::Error`, and `std::fmt::Error` all as `Error`.
Running the rule over this repository produced ten diagnostics, seven of which
were wrong: four in `whisker-macros`, which does not depend on anyhow at all,
two on `fn fmt(..) -> std::fmt::Result`, where `.context()` is not even
expressible, and one on a `serde_json::Error`. The rule's own documentation and
its spec rule both promise it fires only for `anyhow::Error`.

The defect was upstream of the lint. A display string does not carry type
identity, and no amount of string comparison recovers it. The provider now
resolves the error type to the item that defines it and hands the lint a
canonical path — crate name plus module path plus item name — which the lint
compares against a constant. Whether a type is a `Result` or an `Option` goes
through the same route rather than through its rendering.

One more blind spot went with it. `Function::ret_type` gives an `async fn` the
opaque future rather than the awaited type, so `is_result` was false and the
error type was absent for every async function. Every whisker CLI command is
async, so the rule could not see any of them. `Function::async_ret_type` exists
for exactly this and is now used, with `ReturnMode` recording whether a signature
was read directly, awaited, or left opaque, so "async and unprojectable" stays
distinguishable from "async and not a Result".

The lint's own tests had passed throughout, because they attach decorations by
hand and so exercised the rule against decorations the real provider would never
produce. The tests added here run the actual provider over a fixture containing
`anyhow::Result`, `syn::Result`, `std::io::Result`, `std::fmt::Result`, a boxed
error, a generic error, async functions, and a closure returning a different
error type than the function around it, and assert against the whole file rather
than one function at a time — a per-function expectation cannot fail on a rule
that starts firing somewhere else.


martijnarts:

PR #210, opened 2026-08-05
Report the files whisker could not look at

martijnarts:
Three ways of finding nothing were indistinguishable from finding nothing
wrong. A directory whisker could not read was dropped by the walk, so it
vanished from the run. A run that discovered no files at all printed nothing
and exited 0, byte-identical to a clean run, which matters more now that
ignore files can legitimately empty the set. And a file named explicitly was
handed to the Rust parser whatever its extension, so `whisker check Cargo.toml`
parsed TOML as Rust, found nothing, and pronounced it clean.

Walk failures now go through a policy the caller chooses, matching
`--keep-going`: report and carry on, or stop at the first one. Each failure is
classified so the message says what could not be read instead of repeating the
walker's own wording, because "IO error" on its own leaves the user nowhere to
go. Discovering zero files is an error naming both causes, and an explicit
target whose extension has no grammar is refused.

The unreadable-directory tests work through a guard that restores the
directory's mode on drop. Without it, a panic between clearing the mode and
restoring it leaves a directory that cannot be removed, so one assertion
failure would be followed by a second, unrelated one from the temporary
directory's own cleanup.

### DIFF
--- a/crates/whisker/src/discovery.rs
+++ b/crates/whisker/src/discovery.rs
@@ -1,8 +1,14 @@
 use std::path::{Path, PathBuf};
 
 use anyhow::Context as _;
-use ignore::WalkBuilder;
+use ignore::{DirEntry, WalkBuilder};
 use whisker_types::Language;
+
+mod walk_error_policy;
+mod walk_failure;
+
+pub use walk_error_policy::WalkErrorPolicy;
+use walk_failure::WalkFailure;
 
 /// The source files a `whisker check` run inspects
 ///
@@ -14,45 +20,92 @@ use whisker_types::Language;
 #[derive(Debug)]
 pub struct Discovery {
     files: Vec<PathBuf>,
+    errors: Vec<anyhow::Error>,
 }
 
 impl Discovery {
     /// Discovers the files to lint beneath `path`
     ///
     /// A `path` that names one file comes back as-is. Ignore rules do not
-    /// apply to it, because the user already chose the file.
+    /// apply to it, because the user already chose the file. The grammar
+    /// comes from the extension of `path`, not from the name a symlink
+    /// resolves to. Whisker reports the path the user named.
     ///
     /// # Errors
     ///
-    /// Returns an error if `path` cannot be resolved.
+    /// Returns an error if `path` cannot be resolved, if `path` names a file
+    /// with no extension, if `path` names a file written in a language
+    /// whisker has no grammar for, if a configured
+    /// ignore pattern is not valid gitignore syntax, or - under
+    /// [`WalkErrorPolicy::Fail`] - if a directory entry or an ignore file
+    /// cannot be read.
     ///
     /// # Examples
     ///
     /// ```ignore
-    /// let discovery = Discovery::run(Path::new("."))?;
+    /// let discovery = Discovery::run(Path::new("."), WalkErrorPolicy::Fail)?;
     ///
     /// for file in discovery.files() {
     ///     println!("{}", file.display());
     /// }
     /// ```
-    pub fn run(path: &Path) -> anyhow::Result<Self> {
+    pub fn run(path: &Path, on_error: WalkErrorPolicy) -> anyhow::Result<Self> {
         let root = std::fs::canonicalize(path)
             .with_context(|| format!("failed to resolve {}", path.display()))?;
 
         if root.is_file() {
+            let Some(extension) = path.extension() else {
+                anyhow::bail!(
+                    "whisker cannot check {}: it has no file extension, so there is nothing to \
+                     tell whisker which language to parse it as",
+                    path.display()
+                );
+            };
+
+            let extension = extension.to_string_lossy();
+
+            let Some(_language) = Language::from_extension(&extension) else {
+                anyhow::bail!(
+                    "whisker cannot check {}: there is no grammar for `.{extension}` files",
+                    path.display()
+                );
+            };
+
             return Ok(Self {
                 files: vec![path.to_path_buf()],
+                errors: Vec::new(),
             });
         }
 
         let walk = WalkBuilder::new(&root).require_git(false).build();
 
         let mut files = Vec::new();
+        let mut errors = Vec::new();
 
         for entry in walk {
-            let Ok(entry) = entry else {
-                continue;
+            let entry = match entry {
+                Ok(entry) => entry,
+                Err(error) => {
+                    let context = WalkFailure::classify(&error).context();
+                    let error = anyhow::Error::new(error).context(context);
+
+                    match on_error {
+                        WalkErrorPolicy::Fail => return Err(error),
+                        WalkErrorPolicy::ReportAndContinue => {
+                            errors.push(error);
+                            continue;
+                        }
+                    }
+                }
             };
+
+            match attached_error(&entry) {
+                None => {}
+                Some(error) => match on_error {
+                    WalkErrorPolicy::Fail => return Err(error),
+                    WalkErrorPolicy::ReportAndContinue => errors.push(error),
+                },
+            }
 
             let Some(file_type) = entry.file_type() else {
                 continue;
@@ -73,7 +126,7 @@ impl Discovery {
 
         files.sort();
 
-        Ok(Self { files })
+        Ok(Self { files, errors })
     }
 
     /// Returns the discovered files, in a stable order
@@ -86,6 +139,34 @@ impl Discovery {
     pub fn files(&self) -> &[PathBuf] {
         &self.files
     }
+
+    /// Returns the failures the walk survived
+    ///
+    /// The list is always empty under [`WalkErrorPolicy::Fail`], which
+    /// returns the first failure instead of recording it.
+    ///
+    /// # Examples
+    ///
+    /// ```ignore
+    /// for error in discovery.errors() {
+    ///     eprintln!("error: {error:#}");
+    /// }
+    /// ```
+    pub fn errors(&self) -> &[anyhow::Error] {
+        &self.errors
+    }
+}
+
+/// Returns the failure the walker recorded against `entry`, if there was one
+///
+/// An unparsable ignore file inside the tree does not abandon the walk. The
+/// entry arrives intact with the failure attached, and the file's rules are
+/// silently absent. Discovery treats this like an unreadable directory,
+/// because both change the set of linted files.
+fn attached_error(entry: &DirEntry) -> Option<anyhow::Error> {
+    let error = entry.error()?;
+
+    Some(anyhow::Error::msg(error.to_string()).context("failed to read an ignore file"))
 }
 
 /// Rebases `entry` from the resolved `root` onto the `original` argument
@@ -98,9 +179,29 @@ fn rebase(original: &Path, root: &Path, entry: &Path) -> PathBuf {
 
 #[cfg(test)]
 mod tests {
+    #[cfg(unix)]
+    use std::os::unix::fs::PermissionsExt as _;
+
     use tempfile::TempDir;
 
     use super::*;
+
+    /// A directory whose permissions are restored when this value is dropped
+    ///
+    /// A temporary directory left at mode `000` cannot be removed. This
+    /// guard restores the mode even when a test panics, so [`TempDir`]'s
+    /// cleanup still succeeds.
+    #[cfg(unix)]
+    struct Unreadable<'a> {
+        directory: &'a Path,
+    }
+
+    #[cfg(unix)]
+    impl Drop for Unreadable<'_> {
+        fn drop(&mut self) {
+            make_readable(self.directory);
+        }
+    }
 
     /// Panics unless `directory` reads back in unsorted order
     ///
@@ -146,6 +247,40 @@ mod tests {
                     .replace('\\', "/")
             })
             .collect()
+    }
+
+    /// Restores default permissions so the temporary directory can be removed
+    #[cfg(unix)]
+    fn make_readable(directory: &Path) {
+        std::fs::set_permissions(directory, std::fs::Permissions::from_mode(0o755))
+            .expect("permissions should be restored");
+    }
+
+    /// Makes `directory` impossible to read until the returned value is dropped
+    ///
+    /// A privileged process, root in a CI container for example, can still
+    /// read a mode `000` directory. The walk-error tests would then assert
+    /// nothing, so this panics instead of passing silently.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `directory` cannot have its mode cleared, or if it is still
+    /// readable afterwards.
+    #[cfg(unix)]
+    fn make_unreadable(directory: &Path) -> Unreadable<'_> {
+        std::fs::set_permissions(directory, std::fs::Permissions::from_mode(0o000))
+            .expect("permissions should be set");
+
+        let unreadable = Unreadable { directory };
+
+        assert!(
+            std::fs::read_dir(directory).is_err(),
+            "{} is still readable at mode 000, so the walk-error tests cannot exercise anything; \
+             run them without privileges that bypass file permissions",
+            directory.display()
+        );
+
+        unreadable
     }
 
     /// Creates a temporary directory that holds the given relative files
@@ -198,7 +333,8 @@ mod tests {
         ]);
         assert_stored_out_of_order(directory.path());
 
-        let discovery = Discovery::run(directory.path()).expect("discovery should succeed");
+        let discovery = Discovery::run(directory.path(), WalkErrorPolicy::Fail)
+            .expect("discovery should succeed");
 
         assert_eq!(
             discovered(&discovery, directory.path()),
@@ -221,7 +357,8 @@ mod tests {
         std::fs::write(directory.path().join(".ignore"), "generated/\n")
             .expect("ignore file should be written");
 
-        let discovery = Discovery::run(directory.path()).expect("discovery should succeed");
+        let discovery = Discovery::run(directory.path(), WalkErrorPolicy::Fail)
+            .expect("discovery should succeed");
 
         assert_eq!(discovered(&discovery, directory.path()), ["src/main.rs"]);
     }
@@ -236,9 +373,46 @@ mod tests {
             .expect("gitignore should be written");
         let target = directory.path().join("examples");
 
-        let discovery = Discovery::run(&target).expect("discovery should succeed");
+        let discovery =
+            Discovery::run(&target, WalkErrorPolicy::Fail).expect("discovery should succeed");
 
         assert_eq!(discovered(&discovery, &target), ["demo.rs"]);
+    }
+
+    #[test]
+    fn run_with_explicit_extensionless_file_target_returns_error() {
+        let directory = tree(&["LICENSE"]);
+        let target = directory.path().join("LICENSE");
+
+        let error =
+            Discovery::run(&target, WalkErrorPolicy::Fail).expect_err("discovery should fail");
+
+        assert!(
+            format!("{error:#}").contains("no file extension"),
+            "error should explain that the language cannot be told: {error:#}"
+        );
+    }
+
+    /// Pins that a resolved `.rs` target does not rescue an extensionless name
+    ///
+    /// Whisker canonicalizes the path first, and canonicalization resolves
+    /// the symlink. The resolved name must not pick the grammar, because the
+    /// user named `LICENSE`.
+    #[cfg(unix)]
+    #[test]
+    fn run_with_explicit_extensionless_symlink_to_a_rust_file_returns_error() {
+        let directory = tree(&["src/main.rs"]);
+        let link = directory.path().join("LICENSE");
+        std::os::unix::fs::symlink(directory.path().join("src").join("main.rs"), &link)
+            .expect("symlink should be created");
+
+        let error =
+            Discovery::run(&link, WalkErrorPolicy::Fail).expect_err("discovery should fail");
+
+        assert!(
+            format!("{error:#}").contains("no file extension"),
+            "error should explain that the language cannot be told: {error:#}"
+        );
     }
 
     /// A file the user names skips the walk, so no ignore rule reaches it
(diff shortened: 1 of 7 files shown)


crates/whisker/src/discovery/walk_failure.rs
```diff
@@ -0,0 +1,200 @@
...
+                }
+            }
+            Error::Loop {
+                ancestor: _,
+                child: _,
+            }
+            | Error::Io(_)
+            | Error::UnrecognizedFileType(_)
+            | Error::InvalidDefinition => Self::DirectoryEntry,
+        }
```
martijnarts, line 50: It seems pretty bad to just eat these errors? Makes it undebuggable by definition.

martijnarts:
Every doc comment this PR adds was re-checked against three rule sets and rewritten where it failed. 25 blocks changed here.

Violations found here: 2 stated something untrue at this branch, 19 reviewer verdicts from #208, 34 ASD-STE100, 13 AI-writing tells, 7 other style.

|---|---|---|
| `crates/whisker/src/discovery.rs` | Discovery | 40+ word sentences; Figure of speech (worse than useless); Hyphen asides; Essay; stray trailing /// line |
| `crates/whisker/src/discovery.rs` | Discovery::errors | Omitted subject and verb (Always empty under) |
| `crates/whisker/src/discovery.rs` | Discovery::run | 40+ word sentences; Essay; directory-exemption paragraph pins unverified behavior; No configured ignore patterns exist at this branch; Unparsable ignore file errors only under Fail; Hyphe... |
| `crates/whisker/src/discovery.rs` | attached_error | 40+ word sentences; Figure of speech (hanging off it); Hyphen aside |
| `crates/whisker/src/discovery.rs` | rebase | Figure of speech (machine's idea of where that file really lives); 26-word sentence; Wordy for a two-line helper |
| `crates/whisker/src/discovery.rs` | tests::Unreadable | 45-word sentence; Test-helper doc must be short and parseable |
| `crates/whisker/src/discovery.rs` | tests::assert_stored_out_of_order | Unparseable test doc; 40-word sentence; -ing clauses (asserting, quietly passing) |
| `crates/whisker/src/discovery.rs` | tests::make_unreadable | Hyphen aside; 38-word sentence; Figure of speech (bowed out); Inflation (the very failure mode this crate exists to prevent) |
| `crates/whisker/src/discovery.rs` | tests::run_returns_files_in_sorted_order | Unparseable test doc; 40+ word sentences; Figure of speech (says so out loud, visibly its own) |
| `crates/whisker/src/discovery.rs` | tests::run_with_git_exclude_file_excludes_it | Unparseable test doc; 33-word sentence; Figure of speech (fragile part, fall out silently) |
| `crates/whisker/src/discovery.rs` | tests::run_with_unparsable_ignore_file_above_the_root_returns_error | Rather-than; 38-word sentence; Unparseable test doc |
| `crates/whisker/src/discovery/walk_error_policy.rs` | WalkErrorPolicy | Simple enum, wordy doc; 30+ word sentences; Inflation (worst available outcome); --keep-going promise is basically unrelated |
| `crates/whisker/src/discovery/walk_failure.rs` | WalkFailure | 60+ word sentence; Unparseable; Rather-than |
| `crates/whisker/src/discovery/walk_failure.rs` | WalkFailure::classify | 45-word sentence; Implementation detaily before saying what it does; Summary line convoluted (reports having been unable to read) |
| `crates/whisker/tests/check.rs` | UNPARSABLE_IGNORE_FILE | Hyphen aside; 36-word sentence; Figure of speech (cheapest way) |
| `crates/whisker/tests/check.rs` | WARNING_SOURCE | 29-word sentence; Wordy for a constant |
| `crates/whisker/tests/check.rs` | check_package_whose_sources_the_global_gitignore_excludes_fails | Unparseable test doc; 40+ word sentences; Hyphen aside; Spec-promise digression unrelated |
| `crates/whisker/tests/check.rs` | check_package_with_an_unparsable_ignore_file_and_keep_going_reports_it_and_fails | Unparseable test doc; 30+ word sentences; Inflation (would be worse) |
| `crates/whisker/tests/check.rs` | check_package_with_an_unparsable_ignore_file_fails | 30-word sentence; -ing clause (being swapped) |
| `crates/whisker/tests/check.rs` | check_with_keep_going_and_unreadable_file_fails | Unparseable test doc; 34-word sentence; Summary line spans two clauses |
| `crates/whisker/tests/check.rs` | check_with_unreadable_file_and_no_keep_going_fails | Rather-than; 28-word sentence; Summary line is contorted (path that the test is distinguished from) |
| `crates/whisker/tests/check.rs` | package | Unparseable essay; 45-word sentence; Whisker-lint-clean digression basically unrelated |
| `crates/whisker/tests/support/unreadable.rs` | Unreadable | 45-word sentence; Near-duplicate of the discovery.rs guard doc; Inflation (the only thing that makes ... testable at all) |
| `crates/whisker/tests/support/unreadable.rs` | make_readable | Sentence duplicated verbatim in src/discovery.rs tests |
| `crates/whisker/tests/support/unreadable.rs` | make_unreadable | Hyphen aside; Near-duplicate of the discovery.rs helper doc; 38-word sentence; Figure of speech (bowed out) |

- `crates/whisker/src/discovery.rs` &mdash; Discovery::run: No configured ignore patterns exist at this branch

crates/whisker/src/discovery.rs
```diff
@@ -86,6 +136,34 @@ impl Discovery {
     pub fn files(&self) -> &[PathBuf] {
         &self.files
     }
+
+    /// Returns the failures the walk survived
+    ///
+    /// The list is always empty under [`WalkErrorPolicy::Fail`], which
```
martijnarts, line 145: Rewritten. Rules broken:

- Omitted subject and verb (Always empty under)

````rust
    /// Always empty under [`WalkErrorPolicy::Fail`], which returns the first
    /// failure instead of recording it.
````

</details>

File: crates/whisker/src/discovery.rs [hunk 3]
```diff
@@ -86,6 +136,34 @@ impl Discovery {
...
+    /// for error in discovery.errors() {
+    ///     eprintln!("error: {error:#}");
+    /// }
+    /// ```
+    pub fn errors(&self) -> &[anyhow::Error] {
+        &self.errors
+    }
+}
+
+/// Returns the failure the walker recorded against `entry`, if there was one
```
martijnarts, line 160: Rewritten. Rules broken:

- 40+ word sentences
- Figure of speech (hanging off it)
- Hyphen aside

````rust
/// Returns the failure the walker recorded against `entry`, if there was one
///
/// An ignore file whose syntax the walker cannot parse does not abandon the
/// walk and does not arrive as an `Err`: the entry comes back intact with the
/// failure hanging off it, and the rules that file was supposed to express are
/// silently absent. That is the same class of harm as a directory that cannot
/// be read - the set of linted files changes and nothing says so - which is
/// why discovery treats the two the same way.
````

</details>

File: crates/whisker/src/discovery.rs [hunk 4]
```diff
@@ -98,10 +176,30 @@ fn rebase(original: &Path, root: &Path, entry: &Path) -> PathBuf {
...
 #[cfg(test)]
 mod tests {
+    #[cfg(unix)]
+    use std::os::unix::fs::PermissionsExt as _;
+
     use tempfile::TempDir;
 
     use super::*;
 
+    /// A directory whose permissions are restored when this value is dropped
```
martijnarts, line 189: Rewritten. Rules broken:

- 45-word sentence
- Test-helper doc must be short and parseable

````rust
    /// A directory whose permissions are restored when this value is dropped
    ///
    /// The walk-error tests have to survive a panic between clearing a
    /// directory's mode and putting it back: a temporary directory left at
    /// mode `000` cannot be removed, so an assertion failure would be followed
    /// by a second, unrelated failure from [`TempDir`]'s own cleanup, and the
    /// leftover directory would outlive the test run.
````

</details>

File: crates/whisker/src/discovery.rs [hunk 5]
```diff
@@ -148,6 +246,40 @@ mod tests {
...
+    /// Restores default permissions so the temporary directory can be removed
+    #[cfg(unix)]
+    fn make_readable(directory: &Path) {
+        std::fs::set_permissions(directory, std::fs::Permissions::from_mode(0o755))
+            .expect("permissions should be restored");
+    }
+
+    /// Makes `directory` impossible to read until the returned value is dropped
+    ///
+    /// A privileged process, root in a CI container for example, can still
```
martijnarts, line 261: Rewritten. Rules broken:

- Hyphen aside
- 38-word sentence
- Figure of speech (bowed out)
- Inflation (the very failure mode this crate exists to prevent)

````rust
    /// A process with enough privilege - root in a CI container, most often -
    /// can still read a mode `000` directory. A caller that quietly bowed out
    /// there would report a passing test having asserted nothing at all about
    /// the walk-error path, which is the very failure mode this crate exists
    /// to prevent, so this stops loudly instead.
````

</details>

File: crates/whisker/src/discovery.rs [hunk 6]
```diff
@@ -304,11 +471,120 @@ mod tests {
...
             .expect("symlink should be created");
 
-        let discovery = Discovery::run(&link).expect("discovery should succeed");
+        let discovery =
+            Discovery::run(&link, WalkErrorPolicy::Fail).expect("discovery should succeed");
 
         assert_eq!(discovery.files(), vec![link.join("main.rs")]);
     }
 
+    /// Pins that an ignore file is named as one wherever the walker found it
```
martijnarts, line 545: Rewritten. Rules broken:

- Rather-than
- 38-word sentence
- Unparseable test doc

````rust
    /// Pins that an ignore file is named as one wherever the walker found it
    ///
    /// The rules a directory's ancestors impose are collected before the walk
    /// has any entry to hang a failure on, so this fault takes the walker's
    /// generic error path rather than the one every other unparsable ignore
    /// file takes. The user has the same file to fix either way, and being
    /// told to go looking at directory entries instead is worse than being
    /// told nothing.
````

</details>

File: crates/whisker/src/discovery/walk_error_policy.rs [hunk 7]
```diff
@@ -0,0 +1,53 @@
+/// How file discovery reacts to a failure the walk survives
```
martijnarts, line 1: Rewritten. Rules broken:

- Simple enum, wordy doc
- 30+ word sentences
- Inflation (worst available outcome)
- --keep-going promise is basically unrelated

````rust
/// How file discovery reacts to a failure the walk survives
///
/// A directory that cannot be opened, or an entry that cannot be stat'd,
/// silently narrows the set of files a run inspects; an ignore file whose
/// syntax cannot be parsed silently reshapes it. For a linter both are the
/// worst available outcome: it reports success over a scan that never
/// happened. Discovery therefore surfaces every such failure, and mirrors the
/// promise `whisker check --keep-going` already makes about the files it does
/// manage to read.
````

</details>

File: crates/whisker/src/discovery/walk_failure.rs [hunk 8]
```diff
@@ -0,0 +1,193 @@
+use ignore::Error;
+
+/// What a step of the walk was unable to read
```
martijnarts, line 3: Rewritten. Rules broken:

- 60+ word sentence
- Unparseable
- Rather-than

````rust
/// What a step of the walk was unable to read
///
/// The `ignore` crate surfaces an unparsable ignore file two different ways
/// depending on where that file sits. One inside the walk root arrives
/// attached to the directory entry it governs, and one above the walk root is
/// found while the walker collects the rules the root's ancestors impose,
/// before there is any entry to attach it to, so it arrives as a plain failure
/// of the walk instead. The two are the same fault with the same fix, and a
/// user told to look at a directory when the problem is a `.gitignore` one
/// level up has been sent to the wrong place, so the description is decided by
/// what the failure says rather than by which arm of the walk delivered it.
````

</details>

File: crates/whisker/src/discovery/walk_failure.rs [hunk 9]
```diff
@@ -0,0 +1,193 @@
...
+pub enum WalkFailure {
+    /// An entry of the tree could not be read
+    DirectoryEntry,
+
+    /// An ignore file could not be understood
+    IgnoreFile,
+}
+
+impl WalkFailure {
+    /// Returns the kind of failure `error` describes
```
martijnarts, line 19: Rewritten. Rules broken:

- 45-word sentence
- Implementation detaily before saying what it does
- Summary line convoluted (reports having been unable to read)

````rust
    /// Returns what `error` reports having been unable to read
    ///
    /// A glob that does not compile, and a failure carrying the line it
    /// happened on, are only ever produced while parsing an ignore file.
    /// Everything else is treated as a failure of the tree itself, I/O
    /// included: the walker offers no way to tell a directory that cannot be
    /// opened apart from an ignore file inside it that cannot be read, and of
    /// the two the directory is both likelier and the more alarming thing to
    /// point at.
````

</details>

File: crates/whisker/tests/check.rs [hunk 10]
```diff
@@ -4,9 +4,22 @@ use assert_cmd::prelude::*;
...
+#[path = "support/unreadable.rs"]
+mod unreadable;
+
+#[cfg(unix)]
+use unreadable::make_unreadable;
+
 /// Source that trips none of the lints the CLI runs
 const CLEAN_SOURCE: &str = "pub fn answer() -> u32 {\n    42\n}\n";
 
+/// An ignore file the walker cannot parse
```
martijnarts, line 17: Rewritten. Rules broken:

- Hyphen aside
- 36-word sentence
- Figure of speech (cheapest way)

````rust
/// An ignore file the walker cannot parse
///
/// A glob with an unclosed alternate group is the cheapest way to make the
/// walk fail on any platform, which matters because the other way of failing
/// it - a directory nobody may read - does not exist on Windows.
````

</details>

File: crates/whisker/tests/check.rs [hunk 11]
```diff
@@ -79,6 +113,130 @@ fn check_package_directory_succeeds() {
...
+
+    whisker()
+        .arg("check")
+        .arg(package.path())
+        .assert()
+        .failure()
+        .stderr(predicate::str::contains("analyzed no files"));
+}
+
+/// Pins the exclusion source that lives entirely outside the project
```
martijnarts, line 130: Rewritten. Rules broken:

- Unparseable test doc
- 40+ word sentences
- Hyphen aside
- Spec-promise digression unrelated

````rust
/// Pins the exclusion source that lives entirely outside the project
///
/// The user's global gitignore is the last of the four sources the spec
/// promises and the only one whose location whisker never sees: the walker has
/// to read git's own configuration to find it. Nothing inside a checkout can
/// reveal that the lookup stopped happening, so a run would simply start
/// linting whatever that file was hiding, and the first person to notice would
/// be whoever configured it. Pointing `GIT_CONFIG_GLOBAL` at a temporary
/// config keeps the test hermetic - git gives that variable precedence over
/// both `$HOME/.gitconfig` and the XDG config - so it neither reads nor
/// disturbs the developer's own settings.
````

</details>

File: crates/whisker/tests/check.rs [hunk 12]
```diff
@@ -79,6 +113,130 @@ fn check_package_directory_succeeds() {
...
+    whisker()
+        .env("GIT_CONFIG_GLOBAL", &config)
+        .arg("check")
+        .arg(package.path())
+        .assert()
+        .failure()
+        .stderr(predicate::str::contains("analyzed no files"));
+}
+
+/// Pins that `--keep-going` survives a walk error and still fails the run
```
martijnarts, line 156: Rewritten. Rules broken:

- Unparseable test doc
- 30+ word sentences
- Inflation (would be worse)

````rust
/// Pins that `--keep-going` survives a walk error and still fails the run
///
/// Both halves matter and neither is visible from the discovery unit tests,
/// which pin what a policy does once it has been chosen and say nothing about
/// which one the CLI chooses. A run that reported the failure and exited zero
/// would be a linter telling a CI job that source it never read is fine, and a
/// run that reported nothing would be worse.
````

</details>

File: crates/whisker/tests/check.rs [hunk 13]
```diff
@@ -79,6 +113,130 @@ fn check_package_directory_succeeds() {
...
+        .arg(package.path())
+        .assert()
+        .failure()
+        .stderr(predicate::str::contains(
+            "error: failed to read an ignore file",
+        ))
+        .stderr(predicate::str::contains("failed to discover source files").not());
+}
+
+/// Pins that a walk error ends the run when `--keep-going` is not given
```
martijnarts, line 181: Rewritten. Rules broken:

- 30-word sentence
- -ing clause (being swapped)

````rust
/// Pins that a walk error ends the run when `--keep-going` is not given
///
/// The absence of the reported-and-continued message is what distinguishes
/// this from the `--keep-going` case: both exit non-zero, so an exit code
/// alone would not notice the two policies being swapped.
````

</details>

File: crates/whisker/tests/support/unreadable.rs [hunk 14]
```diff
@@ -0,0 +1,49 @@
+use std::os::unix::fs::PermissionsExt as _;
+use std::path::Path;
+
+/// A guard that restores a directory's permissions on drop
```
martijnarts, line 4: Rewritten. Rules broken:

- 45-word sentence
- Near-duplicate of the discovery.rs guard doc
- Inflation (the only thing that makes ... testable at all)

````rust
/// A directory whose permissions are restored when this value is dropped
///
/// The walk-error tests have to survive a panic between clearing a
/// directory's mode and putting it back: a temporary directory left at mode
/// `000` cannot be removed, so an assertion failure would be followed by a
/// second, unrelated failure from `TempDir`'s own cleanup, and the leftover
/// directory would outlive the test run.
///
/// The unit tests in `src/discovery.rs` carry the same guard. An integration
/// test cannot reach into the binary crate it drives, so the duplication buys
/// the only thing that makes the CLI's own handling of a walk error testable
/// at all.
````

</details>

File: crates/whisker/tests/support/unreadable.rs [hunk 15]
```diff
@@ -0,0 +1,49 @@
...
+    directory: &'a Path,
+}
+
+impl Drop for Unreadable<'_> {
+    fn drop(&mut self) {
+        make_readable(self.directory);
+    }
+}
+
+/// Restores default permissions so `TempDir` can remove the directory
```
martijnarts, line 20: Rewritten. Rules broken:

- Sentence duplicated verbatim in src/discovery.rs tests

````rust
/// Restores default permissions so the temporary directory can be removed
````

</details>

File: crates/whisker/tests/support/unreadable.rs [hunk 16]
```diff
@@ -0,0 +1,49 @@
...
+fn make_readable(directory: &Path) {
+    std::fs::set_permissions(directory, std::fs::Permissions::from_mode(0o755))
+        .expect("permissions should be restored");
+}
+
+/// Makes `directory` unreadable until the returned guard is dropped
```
martijnarts, line 26: Rewritten. Rules broken:

- Hyphen aside
- Near-duplicate of the discovery.rs helper doc
- 38-word sentence
- Figure of speech (bowed out)

````rust
/// Makes `directory` impossible to read until the returned value is dropped
///
/// A process with enough privilege - root in a CI container, most often - can
/// still read a mode `000` directory. A caller that quietly bowed out there
/// would report a passing test having asserted nothing at all about the
/// walk-error path, which is the very failure mode this crate exists to
/// prevent, so this stops loudly instead.
///
/// # Panics
///
/// Panics if `directory` cannot have its mode cleared, or if it is still
/// readable afterwards.
````

</details>

**Review by martijnarts — COMMENTED** [review]

Not shown inline: `crates/whisker/src/discovery.rs` Discovery; `crates/whisker/src/discovery.rs` Discovery::run; `crates/whisker/src/discovery.rs` rebase; `crates/whisker/src/discovery.rs` tests::assert_stored_out_of_order; `crates/whisker/src/discovery.rs` tests::run_returns_files_in_sorted_order; `crates/whisker/src/discovery.rs` tests::run_with_git_exclude_file_excludes_it; `crates/whisker/tests/check.rs` WARNING_SOURCE; `crates/whisker/tests/check.rs` package; `crates/whisker/tests/check.rs` check_with_keep_going_and_unreadable_file_fails; `crates/whisker/tests/check.rs` check_with_unreadable_file_and_no_keep_going_fails — those blocks have no added line to anchor to (deleted content, or text a lower branch owns).

File: crates/whisker/src/discovery/walk_failure.rs (hunk 1 above)
Comment (martijnarts) [reply crates/whisker/src/discovery/walk_failure.rs:50]: Nothing eats these errors. `classify` only picks a context string, and the
caller keeps the walker's error as the cause underneath it.

An unreadable directory, which is the arm you flagged, under `--keep-going`:

```text
error: failed to read a directory entry: /tmp/wtest2/locked: IO error for operation on /tmp/wtest2/locked: Permission denied (os error 13)
```

An unparsable `src/.gitignore`, without `--keep-going`:

```text
Error: failed to discover source files

Caused by:
    0: failed to read an ignore file
    1: /tmp/wtest/src/.gitignore: line 1: error parsing glob '{a,b': unclosed alternate group; missing '}' (maybe escape '{' with '[{]'?)
```

Those two come from different lines. A plain walk failure hits
`anyhow::Error::new(error).context(context)`, which keeps the `ignore::Error`
itself as the cause. An ignore file inside the root arrives attached to an
entry that is otherwise fine, so `attached_error` keeps the walker's message
instead. Under `--keep-going`, `check.rs` prints the chain with `{:#}`.
Otherwise the error returns from `check`, and the CLI prints the chain under
`Caused by:`.

You were right that nothing held either line. Every test asserted the static
context string and none asserted the cause, so both stayed green when I
deleted the cause by hand. Two tests now pin it, one per path:

- `run_with_unreadable_directory_and_fail_policy_returns_error` asserts the
  directory's name and `Permission denied`. It is the only test that fails
  when I replace the wrapped error with the bare context string.
- `run_with_unparsable_ignore_file_and_fail_policy_returns_error` asserts the
  ignore file's name and the bad glob. It fails when I drop the message in
  `attached_error`; the CLI tests catch that too.

The two ignore-file tests in `tests/check.rs` also assert the glob text on
stderr. Together they cover both printing routes.

I made two doc changes here. `WalkFailure::context` now says the caller keeps
the walker's error as the cause. The type doc said classification inspects the
failure, not where it arrived. That is not what the code does. `classify` only
ever sees a plain walk failure, because `attached_error` handles the other
route.

crates/whisker/src/discovery.rs
```diff
@@ -47,36 +60,76 @@ impl Discovery {
...
+                    "whisker cannot check {}: there is no grammar for `.{extension}` files",
+                    path.display()
+                );
+            };
+
             return Ok(Self {
                 files: vec![path.to_path_buf()],
+                errors: Vec::new(),
             });
         }
```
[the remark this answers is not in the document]
martijnarts: Fixed. The check now reads `path.extension()`.

Both halves of your example reproduce. Whisker accepted
`LICENSE -> src/main.rs`. Whisker refused `foo.rs -> foo` with
`whisker cannot check foo.rs: it has no file extension, …`. That message
already used `path`, so it contradicted the path it printed.

`path` is the right side to check. The extension is the user's claim about the
language, and `path` is what the user typed. It is also the path whisker reads
and reports. Inside the file branch, the canonical root now only decides that
the target is a file.

The walk cannot settle the question either way. Whisker never asks
`WalkBuilder` to follow symlinks, so the walk drops a symlink entry before any
extension check. An explicit argument is the only place where a link's own
name matters.

Two `#[cfg(unix)]` tests pin the decision:
`run_with_explicit_extensionless_symlink_to_a_rust_file_returns_error` and
`run_with_explicit_rust_symlink_to_an_extensionless_file_returns_the_link`.
Both fail against `root.extension()`. The doc on `Discovery::run` now states
which path picks the grammar, and its `# Errors` list now names the
missing-extension case.

PR #209, opened 2026-08-05
Honor gitignore rules when discovering files

martijnarts:
Whisker walked every file under the path it was given. Pointed at this
repository that meant linting `target/`, hundreds of thousands of generated
files nobody asked to have checked, and it would have kept doing so however
carefully the rules themselves were written.

Discovery now goes through the `ignore` crate, which honors .gitignore,
.ignore, .git/info/exclude, the user's global gitignore, and hidden-file rules:
the same exclusions the rest of a project's tooling already respects. A linter
that reports on generated output is worse than useless, so this is worth doing
on its own terms, before anything is layered on top of it.

The walk still drops the errors it meets, exactly as the one it replaces did.
That is deliberate. Surfacing them changes what a run reports, and that
deserves its own change instead of hiding inside a swap of the walker.

PR #211, opened 2026-08-05
Let a workspace configure what whisker ignores

martijnarts:
Gitignore rules cover build output and anything a checkout already hides, but
they cannot express "this is deliberately violation-laden fixture material for
the lint suite". This repository has two such directories, and git has every
reason to track them.

Projects can now declare a `[workspace.metadata.whisker]` table in the
workspace manifest holding gitignore-syntax patterns, layered on top of the
ignore files. Patterns anchor at the workspace root, so a pattern means the
same thing whether the run starts there or inside a member crate three levels
down.

The table is small on purpose, and a malformed one is a hard error. A typo'd
key that silently ignores nothing is the same class of bug as a linter that
reports a file clean because it could not read it, which is the failure this
whole area keeps producing.

Whisker's own manifest uses the table to exclude the two fixture workspaces.
That exclusion is load-bearing: both sit outside the root workspace, so
rust-analyzer has no data for them, and a later change in this stack turns "no
data" into a hard error. A test loads this repository's own configuration and
asserts no discovered file falls under them, and that whisker's own source
still does, so it cannot pass vacuously.

### DIFF
--- a/crates/whisker/src/config.rs
+++ b/crates/whisker/src/config.rs
@@ -1,0 +1,376 @@
+use std::ffi::OsString;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+use anyhow::Context as _;
+use serde::Deserialize;
+
+mod ignore_pattern;
+
+pub use ignore_pattern::IgnorePattern;
+
+/// Whisker's configuration for a single target project
+///
+/// Whisker reads its configuration from the `[workspace.metadata.whisker]`
+/// table of the target project's workspace manifest, so there is no separate
+/// configuration file. The table currently holds only the [`IgnorePattern`]s
+/// that exclude files from discovery.
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub struct WhiskerConfig {
+    root: PathBuf,
+    ignore: Vec<IgnorePattern>,
+}
+
+impl WhiskerConfig {
+    /// Creates a configuration whose patterns are anchored at `root`
+    ///
+    /// # Examples
+    ///
+    /// ```ignore
+    /// let config = WhiskerConfig::new(
+    ///     PathBuf::from("/src/project"),
+    ///     vec![IgnorePattern::new("examples/")],
+    /// );
+    /// ```
+    pub fn new(root: PathBuf, ignore: Vec<IgnorePattern>) -> Self {
+        Self { root, ignore }
+    }
+
+    /// Loads the configuration that governs `path`
+    ///
+    /// Whisker asks `cargo metadata` for the workspace root and reads the
+    /// metadata table from its output. Cargo is the authority on the
+    /// workspace root: it follows the `package.workspace` key and applies
+    /// workspace inheritance. A directory outside any Cargo project is a
+    /// supported target with an empty configuration.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if `path` cannot be resolved, if `cargo metadata`
+    /// fails, or if the `[workspace.metadata.whisker]` table cannot be read.
+    /// An unrecognized key in the table is an error, because it is most
+    /// likely a typo.
+    pub fn load(path: &Path) -> anyhow::Result<Self> {
+        let root = search_directory(path)?;
+
+        if !has_manifest(&root) {
+            return Ok(Self::new(root, Vec::new()));
+        }
+
+        let cargo = std::env::var_os("CARGO").unwrap_or_else(|| OsString::from("cargo"));
+
+        let output = Command::new(cargo)
+            .args(["metadata", "--no-deps", "--format-version", "1"])
+            .current_dir(&root)
+            .output()
+            .context("failed to run `cargo metadata` to locate the workspace manifest")?;
+
+        anyhow::ensure!(
+            output.status.success(),
+            "failed to read the Cargo workspace containing {}: {}",
+            root.display(),
+            String::from_utf8_lossy(&output.stderr).trim()
+        );
+
+        let CargoMetadata {
+            workspace_root,
+            metadata,
+        } = serde_json::from_slice(&output.stdout)
+            .context("failed to parse the output of `cargo metadata`")?;
+
+        let root = std::fs::canonicalize(&workspace_root).with_context(|| {
+            format!(
+                "failed to resolve the workspace root at {}",
+                workspace_root.display()
+            )
+        })?;
+
+        let Some(metadata) = metadata else {
+            return Ok(Self::new(root, Vec::new()));
+        };
+
+        let Some(table) = metadata.get("whisker") else {
+            return Ok(Self::new(root, Vec::new()));
+        };
+
+        let ConfigTable { ignore } = ConfigTable::deserialize(table).with_context(|| {
+            format!(
+                "failed to read the [workspace.metadata.whisker] table in {}",
+                root.join("Cargo.toml").display()
+            )
+        })?;
+
+        let ignore = ignore.into_iter().map(IgnorePattern::new).collect();
+
+        Ok(Self::new(root, ignore))
+    }
+
+    /// Returns the directory that anchors the ignore patterns
+    ///
+    /// This is the Cargo workspace root when the target is part of a Cargo
+    /// project, and the target directory itself otherwise.
+    ///
+    /// # Examples
+    ///
+    /// ```ignore
+    /// let config = WhiskerConfig::load(Path::new("."))?;
+    ///
+    /// println!("configured from {}", config.root().display());
+    /// ```
+    pub fn root(&self) -> &Path {
+        &self.root
+    }
+
+    /// Returns the patterns that exclude files from discovery
+    ///
+    /// # Examples
+    ///
+    /// ```ignore
+    /// let config = WhiskerConfig::load(Path::new("."))?;
+    ///
+    /// assert!(config.ignore().is_empty());
+    /// ```
+    pub fn ignore(&self) -> &[IgnorePattern] {
+        &self.ignore
+    }
+}
+
+/// The subset of `cargo metadata` output whisker reads
+///
+/// The workspace metadata stays an untyped map so that other tools' tables
+/// pass through untouched. A malformed whisker table then produces an error
+/// that names the whisker table, not the whole document.
+#[derive(Debug, Deserialize)]
+struct CargoMetadata {
+    workspace_root: PathBuf,
+    metadata: Option<serde_json::Map<String, serde_json::Value>>,
+}
+
+/// The `[workspace.metadata.whisker]` table as written in the manifest
+///
+/// The manifest is a system boundary, so this type holds plain strings.
+/// [`WhiskerConfig::load`] turns them into [`IgnorePattern`]s.
+/// Deserialization rejects unknown keys because a silently ignored key looks
+/// exactly like one that works.
+#[derive(Debug, Default, Deserialize)]
+#[serde(deny_unknown_fields)]
+struct ConfigTable {
+    #[serde(default)]
+    ignore: Vec<String>,
+}
+
+/// Returns whether `directory` or any of its ancestors holds a `Cargo.toml`
+fn has_manifest(directory: &Path) -> bool {
+    directory
+        .ancestors()
+        .any(|ancestor| ancestor.join("Cargo.toml").is_file())
+}
+
+/// Returns the absolute directory to ask Cargo about for `path`
+///
+/// For a file target, this is the directory that holds the file.
+///
+/// # Errors
+///
+/// Returns an error if `path` does not exist or cannot be resolved.
+fn search_directory(path: &Path) -> anyhow::Result<PathBuf> {
+    let path = std::fs::canonicalize(path)
+        .with_context(|| format!("failed to resolve {}", path.display()))?;
+
+    if path.is_dir() {
+        return Ok(path);
+    }
+
+    let directory = path.parent().unwrap_or(&path).to_path_buf();
+
+    Ok(directory)
+}
+
+#[cfg(test)]
+mod tests {
+    use tempfile::TempDir;
+
+    use super::*;
+
+    /// A virtual workspace manifest with no members
+    ///
+    /// Members would have to be built to be useful, and every test here is
+    /// about the metadata table, not the crate graph.
+    const EMPTY_WORKSPACE: &str = "[workspace]\nresolver = \"3\"\nmembers = []\n";
+
+    /// Returns the resolved path of a temporary directory
+    ///
+    /// Temporary directories commonly sit behind a symlink, and Cargo reports
+    /// the resolved workspace root, so comparisons have to resolve too.
+    fn resolved(directory: &TempDir) -> PathBuf {
+        std::fs::canonicalize(directory.path()).expect("temporary directory should resolve")
+    }
+
+    /// Creates a temporary Cargo workspace whose manifest is `manifest`
+    fn workspace(manifest: &str) -> TempDir {
+        let directory = tempfile::tempdir().expect("temporary directory should be created");
+        std::fs::write(directory.path().join("Cargo.toml"), manifest)
+            .expect("manifest should be written");
+        directory
+    }
+
+    #[test]
+    fn load_with_empty_ignore_list_returns_an_empty_configuration() {
+        let directory = workspace(&format!(
+            "{EMPTY_WORKSPACE}\n[workspace.metadata.whisker]\nignore = []\n"
+        ));
+
+        let config = WhiskerConfig::load(directory.path()).expect("configuration should load");
+
+        assert!(config.ignore().is_empty());
+    }
+
+    #[test]
+    fn load_with_file_target_uses_the_workspace_root() {
+        let directory = workspace(&format!(
+            "{EMPTY_WORKSPACE}\n[workspace.metadata.whisker]\nignore = [\"examples/\"]\n"
+        ));
+        std::fs::create_dir(directory.path().join("src")).expect("src should be created");
+        let file = directory.path().join("src").join("main.rs");
+        std::fs::write(&file, "fn main() {}").expect("source should be written");
+
+        let config = WhiskerConfig::load(&file).expect("configuration should load");
+
+        assert_eq!(config.root(), resolved(&directory));
+        assert_eq!(config.ignore(), vec![IgnorePattern::new("examples/")]);
+    }
+
+    #[test]
+    fn load_with_ignore_patterns_returns_them_in_order() {
+        let directory = workspace(&format!(
+            "{EMPTY_WORKSPACE}\n[workspace.metadata.whisker]\nignore = [\"examples/\", \"a/b.rs\"]\n"
+        ));
+
+        let config = WhiskerConfig::load(directory.path()).expect("configuration should load");
+
+        assert_eq!(
+            config.ignore(),
+            vec![
+                IgnorePattern::new("examples/"),
+                IgnorePattern::new("a/b.rs")
+            ]
+        );
+    }
+
+    #[test]
+    fn load_with_invalid_manifest_returns_error() {
+        let directory = workspace("[workspace\nthis is not toml\n");
+
+        let error = WhiskerConfig::load(directory.path()).expect_err("configuration should fail");
+
+        assert!(
+            format!("{error:#}").contains("failed to read the Cargo workspace containing"),
+            "error should report that Cargo could not read the manifest: {error:#}"
+        );
+    }
+
+    #[test]
+    fn load_with_malformed_ignore_value_returns_error() {
+        let directory = workspace(&format!(
+            "{EMPTY_WORKSPACE}\n[workspace.metadata.whisker]\nignore = \"examples/\"\n"
+        ));
+
+        let error = WhiskerConfig::load(directory.path()).expect_err("configuration should fail");
+
+        assert!(
+            format!("{error:#}").contains("[workspace.metadata.whisker]"),
+            "error should name the offending table: {error:#}"
+        );
+        assert!(
+            format!("{error:#}").contains("invalid type"),
+            "error should describe the type mismatch: {error:#}"
+        );
+    }
+
+    #[test]
+    fn load_with_no_manifest_returns_an_empty_configuration() {
+        let directory = tempfile::tempdir().expect("temporary directory should be created");
+
+        let config = WhiskerConfig::load(directory.path()).expect("configuration should load");
+
+        assert_eq!(config.root(), resolved(&directory));
+        assert!(config.ignore().is_empty());
+    }
+
+    #[test]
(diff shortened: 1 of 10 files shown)


martijnarts:
Every doc comment this PR adds was re-checked against three rule sets and rewritten where it failed. 14 blocks changed here.

Violations found here: 1 stated something untrue at this branch, 6 reviewer verdicts from #208, 24 ASD-STE100, 6 AI-writing tells.

|---|---|---|
| `crates/whisker/src/config.rs` | WhiskerConfig::ignore | The patterns are not excluded, files that match them are |
| `crates/whisker/src/config.rs` | WhiskerConfig::load | Implementation-detaily essay before saying what the function does; 46-word and 34-word sentences; Em-dash aside in # Errors (- including when it ... -); Figure of speech ("a question Carg... |
| `crates/whisker/src/config.rs` | WhiskerConfig::root | Passive summary ("are resolved against") |
| `crates/whisker/src/config.rs` | fn search_directory | Passive voice in summary and body ("should be asked about", "is answered with"); 25-word sentence at the cap with an -ing clause ("holding it") |
| `crates/whisker/src/config.rs` | struct CargoMetadata | 43-word sentence; Passive voice ("is kept") |
| `crates/whisker/src/config.rs` | struct ConfigTable | Em-dash aside (- a project believing ... -); 42-word sentence; Significance inflation ("the kind of quiet wrongness this project exists to prevent") |
| `crates/whisker/src/config.rs` | struct WhiskerConfig | Essay about anticipated future config keys is basically unrelated to the type; 33-word and 40-word sentences; Em-dash aside (- per-rule severity and per-rule enablement -) |
| `crates/whisker/src/config/ignore_pattern.rs` | IgnorePattern::new | 36-word sentence; Passive voice ("can be reported", "is anchored to") |
| `crates/whisker/src/config/ignore_pattern.rs` | struct IgnorePattern | Codestyle essay (why gitignore syntax was chosen) on a simple newtype; 40-word and 42-word sentences; -ing clauses ("Inheriting gitignore syntax also means inheriting"); Passive voice ("i... |
| `crates/whisker/src/discovery.rs` | Discovery::run (block modified by this branch) | Unparseable essay before the behavior ("refusing at that point would look like whisker silently doing nothing"); 43-word and 41-word sentences; Em-dash aside in # Errors (- under [`WalkEr... |
| `crates/whisker/src/discovery.rs` | const EMPTY_WORKSPACE (tests) | Near-duplicate of the config.rs EMPTY_WORKSPACE doc; Idiom ("never enters into it") and -ing clause ("the tests reaching for a manifest") |
| `crates/whisker/src/discovery.rs` | fn build_excludes | 46-word and 38-word sentences; Passive voice ("is anchored", "is built"); -ing clauses ("being scanned", "naming the offending pattern") |
| `crates/whisker/src/discovery.rs` | struct Discovery (block modified by this branch) | Em-dash aside around the exclusion list; 45-word and 41-word sentences; Figure of speech ("worse than useless", "hands on", "instead of thrown away") |
| `crates/whisker/src/discovery.rs` | test run_over_whiskers_own_workspace_excludes_its_fixture_projects | Test docs must be one or two parseable sentences, not a five-sentence essay; 42-word sentence; Figure of speech ("how a project learns to ignore its own linter", "files that exist to be w... |

- `crates/whisker/src/config.rs` &mdash; WhiskerConfig::ignore: The patterns are not excluded, files that match them are

crates/whisker/src/config.rs
```diff
@@ -0,0 +1,376 @@
...
+use std::process::Command;
+
+use anyhow::Context as _;
+use serde::Deserialize;
+
+mod ignore_pattern;
+
+pub use ignore_pattern::IgnorePattern;
+
+/// Whisker's configuration for a single target project
```
martijnarts, line 12: Rewritten. Rules broken:

- Essay about anticipated future config keys is basically unrelated to the type
- 33-word and 40-word sentences
- Em-dash aside (- per-rule severity and per-rule enablement -)

````rust
/// Whisker's configuration for a single target project
///
/// Whisker is configured from the `[workspace.metadata.whisker]` table of the
/// target project's Cargo workspace manifest. Keeping configuration there
/// means there is no new file to discover, no new file format to learn, and no
/// question about which of several config files won: the workspace manifest is
/// already the one file every Cargo-aware tool agrees on.
///
/// The table is deliberately small. It carries only [`IgnorePattern`]s today,
/// and the richer configuration the specification anticipates - per-rule
/// severity and per-rule enablement - is meant to arrive as additional keys
/// beside `ignore`, each becoming another field on this type and another
/// accessor. Nothing about the shape of this type has to change to accommodate
/// them, and callers that only care about discovery keep reading
/// [`WhiskerConfig::ignore`] unchanged.
````

</details>

**Review by martijnarts — COMMENTED** [review]

File: crates/whisker/src/config.rs [hunk 2]
```diff
@@ -0,0 +1,376 @@
...
+    /// let config = WhiskerConfig::new(
+    ///     PathBuf::from("/src/project"),
+    ///     vec![IgnorePattern::new("examples/")],
+    /// );
+    /// ```
+    pub fn new(root: PathBuf, ignore: Vec<IgnorePattern>) -> Self {
+        Self { root, ignore }
+    }
+
+    /// Loads the configuration that governs `path`
```
martijnarts, line 39: Rewritten. Rules broken:

- Implementation-detaily essay before saying what the function does
- 46-word and 34-word sentences
- Em-dash aside in # Errors (- including when it ... -)
- Figure of speech ("a question Cargo already answers", quoted hypothetical bug report)

````rust
    /// Loads the configuration that governs `path`
    ///
    /// Whisker asks Cargo where the workspace is instead of searching for a
    /// `[workspace]` table itself. `cargo metadata` is the authority on both
    /// the workspace root and its metadata table: it follows the
    /// `package.workspace` key that lets a member point at a root which is not
    /// one of its ancestors, it applies workspace inheritance, and it reports
    /// the same root that rust-analyzer will later load. A hand-rolled upward
    /// search would be a second, subtly different answer to a question Cargo
    /// already answers, and the two disagreeing is exactly the sort of bug
    /// that surfaces as "my ignore pattern does nothing".
    ///
    /// Whisker does perform its own upward search, but only to decide whether
    /// there is a manifest to ask Cargo about at all. A directory that is not
    /// part of any Cargo project is a supported target that simply has no
    /// configuration, not a Cargo error the user has to interpret.
    ///
    /// # Errors
    ///
    /// Returns an error if `path` cannot be resolved, if `cargo metadata`
    /// cannot be run or reports a failure, or if the
    /// `[workspace.metadata.whisker]` table cannot be read - including when it
    /// contains a key whisker does not recognize, which is far more likely to
    /// be a typo than a deliberate annotation for some other tool.
````

</details>

File: crates/whisker/src/config.rs [hunk 3]
```diff
@@ -0,0 +1,376 @@
...
+                root.join("Cargo.toml").display()
+            )
+        })?;
+
+        let ignore = ignore.into_iter().map(IgnorePattern::new).collect();
+
+        Ok(Self::new(root, ignore))
+    }
+
+    /// Returns the directory that anchors the ignore patterns
```
martijnarts, line 108: Rewritten. Rules broken:

- Passive summary ("are resolved against")

````rust
    /// Returns the directory the ignore patterns are resolved against
    ///
    /// This is the Cargo workspace root when the target is part of a Cargo
    /// project, and the target directory itself otherwise.
    ///
    /// # Examples
    ///
    /// ```ignore
    /// let config = WhiskerConfig::load(Path::new("."))?;
    ///
    /// println!("configured from {}", config.root().display());
    /// ```
````

</details>

File: crates/whisker/src/config.rs [hunk 4]
```diff
@@ -0,0 +1,376 @@
...
+    /// ```ignore
+    /// let config = WhiskerConfig::load(Path::new("."))?;
+    ///
+    /// println!("configured from {}", config.root().display());
+    /// ```
+    pub fn root(&self) -> &Path {
+        &self.root
+    }
+
+    /// Returns the patterns that exclude files from discovery
```
martijnarts, line 124: Rewritten. Rules broken:

- The patterns are not excluded, files that match them are

````rust
    /// Returns the patterns excluded from file discovery
    ///
    /// # Examples
    ///
    /// ```ignore
    /// let config = WhiskerConfig::load(Path::new("."))?;
    ///
    /// assert!(config.ignore().is_empty());
    /// ```
````

</details>

File: crates/whisker/src/config.rs [hunk 5]
```diff
@@ -0,0 +1,376 @@
...
+    /// let config = WhiskerConfig::load(Path::new("."))?;
+    ///
+    /// assert!(config.ignore().is_empty());
+    /// ```
+    pub fn ignore(&self) -> &[IgnorePattern] {
+        &self.ignore
+    }
+}
+
+/// The subset of `cargo metadata` output whisker reads
```
martijnarts, line 138: Rewritten. Rules broken:

- 43-word sentence
- Passive voice ("is kept")

````rust
/// The subset of `cargo metadata` output whisker reads
///
/// The workspace metadata is kept as an untyped map, not a struct, so that
/// other tools' tables under `[workspace.metadata]` pass through untouched,
/// and so that a malformed whisker table produces an error naming the whisker
/// table instead of the whole metadata document.
````

</details>

File: crates/whisker/src/config.rs [hunk 6]
```diff
@@ -0,0 +1,376 @@
...
+/// The workspace metadata stays an untyped map so that other tools' tables
+/// pass through untouched. A malformed whisker table then produces an error
+/// that names the whisker table, not the whole document.
+#[derive(Debug, Deserialize)]
+struct CargoMetadata {
+    workspace_root: PathBuf,
+    metadata: Option<serde_json::Map<String, serde_json::Value>>,
+}
+
+/// The `[workspace.metadata.whisker]` table as written in the manifest
```
martijnarts, line 149: Rewritten. Rules broken:

- Em-dash aside (- a project believing ... -)
- 42-word sentence
- Significance inflation ("the kind of quiet wrongness this project exists to prevent")

````rust
/// The `[workspace.metadata.whisker]` table as written in the manifest
///
/// This is the wire shape, which is why it holds plain strings: the manifest
/// is a system boundary, and turning what it says into [`IgnorePattern`]s is
/// the job of [`WhiskerConfig::load`].
///
/// Unknown keys are rejected. A configuration key that whisker silently
/// ignores looks exactly like a configuration key that works, and the failure
/// mode - a project believing it excluded a directory that is still being
/// linted - is the kind of quiet wrongness this project exists to prevent.
````

</details>

File: crates/whisker/src/config.rs [hunk 7]
```diff
@@ -0,0 +1,376 @@
...
+}
+
+/// Returns whether `directory` or any of its ancestors holds a `Cargo.toml`
+fn has_manifest(directory: &Path) -> bool {
+    directory
+        .ancestors()
+        .any(|ancestor| ancestor.join("Cargo.toml").is_file())
+}
+
+/// Returns the absolute directory to ask Cargo about for `path`
```
martijnarts, line 169: Rewritten. Rules broken:

- Passive voice in summary and body ("should be asked about", "is answered with")
- 25-word sentence at the cap with an -ing clause ("holding it")

````rust
/// Returns the absolute directory Cargo should be asked about for `path`
///
/// A file target is answered with the directory holding it, since a file has
/// no manifest of its own but the crate around it does.
///
/// # Errors
///
/// Returns an error if `path` does not exist or cannot be resolved.
````

</details>

File: crates/whisker/src/config/ignore_pattern.rs [hunk 8]
```diff
@@ -0,0 +1,92 @@
+/// A gitignore-syntax pattern that excludes files from a `whisker check` run
```
martijnarts, line 1: Rewritten. Rules broken:

- Codestyle essay (why gitignore syntax was chosen) on a simple newtype
- 40-word and 42-word sentences
- -ing clauses ("Inheriting gitignore syntax also means inheriting")
- Passive voice ("is matched against", "whisker was pointed at")

````rust
/// A pattern whose matches are kept out of a `whisker check` run
///
/// The syntax is gitignore's, not a bespoke glob dialect. Projects
/// already write and read gitignore patterns, whisker already honors the
/// `.gitignore` files it finds, and inventing a second pattern language would
/// mean two sets of rules to reason about when a file unexpectedly is or is
/// not linted. A trailing slash therefore restricts a pattern to directories,
/// `**` crosses directory boundaries, and a leading `!` re-includes something
/// an earlier pattern excluded. Inheriting gitignore syntax also means
/// inheriting its one surprise: a `!` cannot re-include a file whose parent
/// directory is itself excluded, because the walk never descends into that
/// directory to reach the file.
///
/// A pattern is matched against the Cargo workspace root instead of the
/// directory whisker was pointed at, so it means the same thing whether
/// the run starts at the workspace root or inside a member crate. Within that
/// root it anchors the way gitignore anchors: `crates/app/generated/` names
/// one directory relative to the root, a bare `examples/` matches a directory
/// of that name at any depth beneath it, and `/examples/` restricts it to the
/// root.
///
/// # Examples
///
/// ```ignore
/// let pattern = IgnorePattern::new("examples/");
///
/// assert_eq!(pattern.as_str(), "examples/");
/// ```
````

</details>

File: crates/whisker/src/config/ignore_pattern.rs [hunk 9]
```diff
@@ -0,0 +1,92 @@
...
+/// ```ignore
+/// let pattern = IgnorePattern::new("examples/");
+///
+/// assert_eq!(pattern.as_str(), "examples/");
+/// ```
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub struct IgnorePattern(String);
+
+impl IgnorePattern {
+    /// Creates a pattern from its gitignore-syntax source
```
martijnarts, line 24: Rewritten. Rules broken:

- 36-word sentence
- Passive voice ("can be reported", "is anchored to")

````rust
    /// Creates a pattern from its gitignore-syntax source
    ///
    /// The source is not validated here. A pattern is only compiled once the
    /// workspace root it is anchored to is known, which is where a syntax
    /// error can be reported against the manifest that introduced it.
    ///
    /// # Examples
    ///
    /// ```ignore
    /// let pattern = IgnorePattern::new("target/");
    /// ```
````

</details>

File: crates/whisker/src/discovery.rs [hunk 10]
```diff
@@ -1,22 +1,36 @@
...
+use crate::config::WhiskerConfig;
+
 mod walk_error_policy;
 mod walk_failure;
 
 pub use walk_error_policy::WalkErrorPolicy;
 use walk_failure::WalkFailure;
 
-/// The source files a `whisker check` run inspects
+/// The source files a `whisker check` run will inspect
```
martijnarts, line 16: Rewritten. Rules broken:

- Em-dash aside around the exclusion list
- 45-word and 41-word sentences
- Figure of speech ("worse than useless", "hands on", "instead of thrown away")

````rust
/// The source files a `whisker check` run will inspect
///
/// Discovery is deliberately conservative about what it hands on. A linter
/// that reports on generated output is worse than useless, so the walk honors
/// the same exclusions the project's own tooling honors - `.gitignore`,
/// `.ignore`, `.git/info/exclude`, the user's global gitignore, and hidden
/// files - before applying whisker's own [`IgnorePattern`]s on top.
///
/// Those files are honored whether or not there is a repository around them,
/// which is where whisker departs from what `ripgrep` does by default. A
/// `.gitignore` inside an exported tarball, a vendored dependency, or a
/// checkout whose `.git` directory was stripped still describes which files
/// the project generates, and that description does not stop being true
/// because the repository is gone.
///
/// Failures are carried alongside the files instead of thrown away. An
/// unreadable directory quietly shrinks the scan and an unparsable ignore file
/// quietly reshapes it, either of which would let a run report success over
/// source it never opened, so the caller always gets the chance to act on
/// them.
///
/// [`IgnorePattern`]: crate::config::IgnorePattern
````

</details>

File: crates/whisker/src/discovery.rs [hunk 11]
```diff
@@ -26,27 +40,39 @@ pub struct Discovery {
 impl Discovery {
     /// Discovers the files to lint beneath `path`
     ///
-    /// A `path` that names one file comes back as-is. Ignore rules do not
-    /// apply to it, because the user already chose the file.
+    /// When `path` names a single file, the run returns it as-is, and no
```
martijnarts, line 43: Rewritten. Rules broken:

- Unparseable essay before the behavior ("refusing at that point would look like whisker silently doing nothing")
- 43-word and 41-word sentences
- Em-dash aside in # Errors (- under [`WalkErrorPolicy::Fail`] -)
- Passive voice throughout ("is returned", "is walked", "is not reapplied")

````rust
    /// Discovers the files to lint beneath `path`
    ///
    /// A `path` naming a single file is returned as-is, provided whisker has a
    /// grammar for it. Ignore rules describe what a broad sweep should skip,
    /// and a user who types a file's name has already made the decision the
    /// rules exist to automate; refusing at that point would look like whisker
    /// silently doing nothing. Whether whisker can parse the file at all is a
    /// different question, and not one the user can overrule: handing a
    /// `Cargo.toml` to the Rust grammar produces a tree nothing matches, so it
    /// would be reported as clean without ever having been understood.
    ///
    /// The same exemption applies to a directory named on the command line: it
    /// is walked even when a pattern would have excluded it, and the pattern
    /// that excluded it is not reapplied to the files inside, since asking for
    /// a directory is asking for what it contains. Patterns that match
    /// something further down still prune it.
    ///
    /// # Errors
    ///
    /// Returns an error if `path` cannot be resolved, if `path` names a file
    /// written in a language whisker has no grammar for, if a configured
    /// ignore pattern is not valid gitignore syntax, or - under
    /// [`WalkErrorPolicy::Fail`] - if a directory entry or an ignore file
    /// cannot be read.
    ///
    /// # Examples
    ///
    /// ```ignore
    /// let config = WhiskerConfig::load(Path::new("."))?;
    /// let discovery = Discovery::run(Path::new("."), &config, WalkErrorPolicy::Fail)?;
    ///
    /// for file in discovery.files() {
    ///     println!("{}", file.display());
    /// }
    /// ```
````

</details>

File: crates/whisker/src/discovery.rs [hunk 12]
```diff
@@ -166,6 +207,32 @@ fn attached_error(entry: &DirEntry) -> Option<anyhow::Error> {
     Some(anyhow::Error::msg(error.to_string()).context("failed to read an ignore file"))
 }
 
+/// Compiles the configured ignore patterns into a matcher
```
martijnarts, line 214: Rewritten. Rules broken:

- 46-word and 38-word sentences
- Passive voice ("is anchored", "is built")
- -ing clauses ("being scanned", "naming the offending pattern")

````rust
/// Compiles the configured ignore patterns into a matcher
///
/// The matcher is anchored at the configured root, not at the directory
/// being scanned, which is why it is built from the configuration: `examples/`
/// has to mean the same thing whether the run starts at the workspace root or
/// inside a member crate three levels down. Within that root the patterns
/// behave exactly as they would in a `.gitignore` written there, so one
/// containing an interior slash names a path relative to the root while one
/// without it matches at any depth.
///
/// # Errors
///
/// Returns an error naming the offending pattern if it is not valid gitignore
/// syntax.
````

</details>

File: crates/whisker/src/discovery.rs [hunk 13]
```diff
@@ -182,6 +249,13 @@ mod tests {
     use tempfile::TempDir;
 
     use super::*;
+    use crate::config::IgnorePattern;
+
+    /// A virtual workspace manifest with no members
```
martijnarts, line 257: Rewritten. Rules broken:

- Near-duplicate of the config.rs EMPTY_WORKSPACE doc
- Idiom ("never enters into it") and -ing clause ("the tests reaching for a manifest")

````rust
    /// A virtual workspace manifest with no members
    ///
    /// Members would have to be buildable to be useful; the tests reaching
    /// for a manifest care about the metadata table, and the crate graph
    /// never enters into it.
````

</details>

File: crates/whisker/src/discovery.rs [hunk 14]
```diff
@@ -313,6 +411,46 @@ mod tests {
         assert_eq!(path, Path::new("./src/main.rs"));
     }
 
+    /// Pins the exclusions whisker's own manifest declares
```
martijnarts, line 417: Rewritten. Rules broken:

- Test docs must be one or two parseable sentences, not a five-sentence essay
- 42-word sentence
- Figure of speech ("how a project learns to ignore its own linter", "files that exist to be wrong")

````rust
    /// Pins the exclusions whisker's own manifest declares
    ///
    /// Whisker is meant to be run over whisker, and the fixture projects it
    /// keeps around are deliberately full of the violations its rules look
    /// for. They also sit outside the workspace's crate graph, so nothing can
    /// decorate them. A pattern that stopped matching would therefore not fail
    /// loudly; it would fill an otherwise clean run with diagnostics about
    /// files that exist to be wrong, which is how a project learns to ignore
    /// its own linter. The synthetic tests above prove the matching rules, and
    /// this one proves they were spelled correctly here.
````

</details>

File: crates/whisker/src/discovery.rs [hunk 15]
```diff
@@ -100,7 +109,22 @@ impl Discovery {
             });
```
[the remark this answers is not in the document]
martijnarts: You are right, and this is now fixed. `Discovery::run` compiles the configured
patterns immediately after it resolves the target path, before the branch that
handles an explicit file. An invalid pattern is therefore an error for a file
target and a directory target alike.

File selection is unchanged. Only the walk's `filter_entry` consults the
compiled matcher. An explicit file target still comes back as-is, and no
pattern touches it.

The new test `run_with_explicit_file_target_and_invalid_pattern_returns_error`
runs the same broken pattern twice, once against a file target and once
against the tree root. Both runs produce the same error string.

## C2 — crates/whisker/Cargo.toml — dependency handover from #209
VERDICT: fixed
This branch needs both `serde` and `serde_json`, and both are present in
`crates/whisker/Cargo.toml`. `WhiskerConfig::load` shells out to
`cargo metadata` and reads the JSON with `serde_json::from_slice`. The target
type `CargoMetadata` holds the workspace metadata as an
`Option<serde_json::Map<String, serde_json::Value>>`, and `ConfigTable`
derives `serde::Deserialize` with `deny_unknown_fields`.
`crates/whisker/src/config.rs` is the only source file in the crate that
touches either dependency.

The same commit that adds the feature also adds the README section on
`[workspace.metadata.whisker]`.

## Follow-up — crates/whisker/src/discovery.rs — doc damage from 244632a
VERDICT: fixed
Commit 244632a left two doc blocks with redundant text. In both, a rebase
reinstated pre-review wording on top of the text that replaced it. Above
`fn rebase` that is a second summary line over the existing one. Above the
test `run_returns_files_in_sorted_order` it is a long paragraph over the short
one. I deleted both additions, so each block again reads exactly as it does on
the parent branch.

The pushed parent `origin/dogfood/03-discovery-errors` still carries both old
wordings, which is where the rebase picked them up.

PR #212, opened 2026-08-05
Add the vocabulary for describing coverage gaps

martijnarts:
A decoration provider currently has one way to say "I have nothing for this
file": return success. The pipeline cannot distinguish that from "I decorated
it", so it walks an undecorated tree and the rules that need semantic
information report the file clean.

Fixing that needs somewhere to put the answer, and the answer is not a boolean.
A provider that declines has a reason. The reason determines what the user
should do about it. And a run over many files wants to print each distinct
remedy once, not repeat it under every file. This change adds the types that
carry all of it: `Coverage` for the decision, `CoverageGap` for the reason with
the remedy it implies, `ProviderName` so a gap says who declined, and
`UncoveredFile` to gather the gaps reported against one file.

Nothing uses them yet. They land on their own so the change that rewrites the
provider contract is a change to the contract, not the contract plus the
invention of everything it talks about.


crates/whisker-types/src/coverage_gap.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// loaded, so nothing the provider knows can apply to it
+    ///
+    /// The remedy is to point the provider at a workspace that contains the
+    /// file, which is why this is distinct from [`CoverageGap::Unreachable`]:
+    /// a file the user cannot move into the loaded workspace needs a
+    /// different run, not a different module tree.
+    OutsideWorkspace {
+        /// The workspace root the provider loaded
+        root: Arc<Path>,
+    },
```
martijnarts: Workspace I think is very Rust-specific language here? Not sure that applies to other languages. Maybe we should rename?

crates/whisker-types/src/decorated_tree.rs
```diff
@@ -29,12 +29,30 @@ impl DecoratedTree {
...
+    ///
+    /// A [`DecorationProvider`] cannot reach this: it is handed the tree
+    /// immutably and returns its decorations instead, which the pipeline
+    /// applies with [`DecoratedTree::merge_decorations`]. This escape hatch
+    /// exists so a test can stage a decoration without loading a toolchain.
+    ///
+    /// [`DecorationProvider`]: crate::DecorationProvider
     pub fn decorations_mut(&mut self) -> &mut DecorationMap {
         &mut self.decorations
     }
```
martijnarts: if this is only relevant for tests then maybe it should be `#[cfg(test)]`? Does that work?

crates/whisker-types/src/provider_name.rs
```diff
@@ -0,0 +1,79 @@
...
+///
+/// ```
+/// use whisker_types::ProviderName;
+///
+/// assert_eq!(ProviderName("rust").to_string(), "rust");
+/// ```
+///
+/// [`RuleId`]: crate::RuleId
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub struct ProviderName(pub &'static str);
```
martijnarts, line 16: Create a GH issue for disallowing this, this should use the getset crate instead if allowed.

Any `pub` items actually.

martijnarts:
Every doc comment this PR adds was re-checked against three rule sets and rewritten where it failed. 13 blocks changed here.

Violations found here: 3 stated something untrue at this branch, 9 reviewer verdicts from #208, 14 ASD-STE100, 3 AI-writing tells.

|---|---|---|
| `crates/whisker-types/src/coverage.rs` | enum Coverage | At this branch DecorationProvider::decorate takes &mut DecoratedTree and writes through it, so 'decorations travel back by value' is false; 'no one thought that' design essay (declining-p... |
| `crates/whisker-types/src/coverage_gap.rs` | CoverageGap::ExcludedByToolchain | Implementation detail ('holds the file's identity but deliberately not its contents') without saying the plain WHAT first |
| `crates/whisker-types/src/coverage_gap.rs` | CoverageGap::OutsideWorkspace | Essay unrelated to the variant (remedy rationale belongs to help(), cross-variant comparison essay); 37-word sentence ('The remedy is to point the provider...') |
| `crates/whisker-types/src/coverage_gap.rs` | CoverageGap::Unreachable | 27-word sentence ('The file lives under the workspace root but...'); 37-word sentence ('This covers both a file the toolchain interned...'); Implementation detail ('interned', 'attaching ... |
| `crates/whisker-types/src/coverage_gap.rs` | enum CoverageGap | Simple enum with wordy explanation; 32-word sentence ('The variants are distinct because...'); Figure of speech ('the only thing standing between the user and a linter that quietly analyz... |
| `crates/whisker-types/src/coverage_gap.rs` | fn CoverageGap::help | 33-word sentence ('The remedy is separate from Display because...') |
| `crates/whisker-types/src/decorated_tree.rs` | fn DecoratedTree::decorations_mut | At this branch DecorationProvider::decorate receives &mut DecoratedTree and whisker-rust's provider calls decorations_mut, so 'a DecorationProvider cannot reach this' is false; 'no one th... |
| `crates/whisker-types/src/decorated_tree.rs` | fn DecoratedTree::merge_decorations | At this branch DecorationProvider receives the tree mutably and no pipeline calls merge_decorations, so 'the only path available to a DecorationProvider' is false; 30-word sentence |
| `crates/whisker-types/src/decoration_map.rs` | fn DecorationMap::merge | 38-word sentence ('Entries already present win: ... the provider consulted first is the one lint rules see'); Explains pipeline consequences unrelated to this method's own behavior |
| `crates/whisker-types/src/provider_name.rs` | struct ProviderName | 27-word sentence; Wordy rationale for a simple newtype ('the same shape RuleId uses for the same reason') |
| `crates/whisker-types/src/uncovered_file.rs` | fn UncoveredFile::gaps | Omitted subject and verb ('Empty when...'); Filler ('at all') |
| `crates/whisker-types/src/uncovered_file.rs` | fn UncoveredFile::new | Filler ('at all') |
| `crates/whisker-types/src/uncovered_file.rs` | struct UncoveredFile | 26-word sentence with omitted subject ('Carried through...'); Essay ('The alternative, collapsing ... is the failure this type exists to prevent'); Figure of speech (quoted inner voice 'I... |

- `crates/whisker-types/src/coverage.rs` &mdash; enum Coverage: At this branch DecorationProvider::decorate takes &mut DecoratedTree and writes through it, so 'decorations travel back by value' is false
- `crates/whisker-types/src/decorated_tree.rs` &mdash; fn DecoratedTree::decorations_mut: At this branch DecorationProvider::decorate receives &mut DecoratedTree and whisker-rust's provider calls decorations_mut, so 'a DecorationProvider cannot reach this' is false
- `crates/whisker-types/src/decorated_tree.rs` &mdash; fn DecoratedTree::merge_decorations: At this branch DecorationProvider receives the tree mutably and no pipeline calls merge_decorations, so 'the only path available to a DecorationProvider' is false

crates/whisker-types/src/coverage.rs
```diff
@@ -0,0 +1,39 @@
+use crate::{CoverageGap, DecorationMap};
+
+/// The verdict from one decoration provider for one file
```
martijnarts, line 3: Rewritten. Rules broken:

- At this branch DecorationProvider::decorate takes &mut DecoratedTree and writes through it, so 'decorations travel back by value' is false
- 'no one thought that' design essay (declining-provider harmlessness)
- Figure of speech ('a provider that says so is telling the truth')
- 30-word sentence ('That is what makes a declining provider harmless...')

````rust
/// The result of offering one file to one decoration provider
///
/// Decorations travel back by value instead of being written through a
/// `&mut DecoratedTree`. That is what makes a declining provider harmless:
/// it never holds a handle it could mutate, so "declined" and "left partial
/// decorations behind" cannot both be true.
///
/// An empty [`DecorationMap`] inside [`Coverage::Covered`] is a legitimate
/// verdict. A file containing only `use` statements has nothing to
/// decorate, and a provider that says so is telling the truth. Coverage is
/// a claim about the *file*, not about individual nodes.
````

</details>

File: crates/whisker-types/src/coverage_gap.rs [hunk 5]
```diff
@@ -0,0 +1,222 @@
+use std::path::Path;
+use std::sync::Arc;
+
+/// Why a decoration provider declined to decorate a file
```
martijnarts, line 4: Rewritten. Rules broken:

- Simple enum with wordy explanation
- 32-word sentence ('The variants are distinct because...')
- Figure of speech ('the only thing standing between the user and a linter that quietly analyzes nothing')
- Significance inflation

````rust
/// Why a decoration provider declined to decorate a file
///
/// A gap is not a provider failure; it is a provider correctly reporting
/// that it has nothing truthful to say about a file. The variants are
/// distinct because each implies a different fix for the user, and the
/// message they produce is the only thing standing between the user and a
/// linter that quietly analyzes nothing.
///
/// The enum is deliberately not `#[non_exhaustive]`: a new failure mode
/// must break every match that renders a gap, not fall into a wildcard arm
/// that says nothing actionable.
````

</details>

File: crates/whisker-types/src/coverage_gap.rs [hunk 6]
```diff
@@ -0,0 +1,222 @@
...
+///
+/// A gap is not a failure: the provider reports why it cannot analyze
+/// the file. Each variant implies a different remedy, which
+/// [`CoverageGap::help`] returns.
+///
+/// The enum is not `#[non_exhaustive]`: a new variant must break every
+/// match that renders a gap.
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub enum CoverageGap {
+    /// The file is outside the workspace root the provider loaded
```
martijnarts: Rewritten. Rules broken:

- Essay unrelated to the variant (remedy rationale belongs to help(), cross-variant comparison essay)
- 37-word sentence ('The remedy is to point the provider...')

````rust
    /// The file lives outside the workspace root directory the provider
    /// loaded, so nothing the provider knows can apply to it
    ///
    /// The remedy is to point the provider at a workspace that contains the
    /// file, which is why this is distinct from [`CoverageGap::Unreachable`]:
    /// a file the user cannot move into the loaded workspace needs a
    /// different run, not a different module tree.
````

</details>

File: crates/whisker-types/src/coverage_gap.rs [hunk 7]
```diff
@@ -0,0 +1,222 @@
...
+    OutsideWorkspace {
+        /// The workspace root the provider loaded
+        root: Arc<Path>,
+    },
+
+    /// The file is under the workspace root, but no crate in the
```
martijnarts: Rewritten. Rules broken:

- 27-word sentence ('The file lives under the workspace root but...')
- 37-word sentence ('This covers both a file the toolchain interned...')
- Implementation detail ('interned', 'attaching to a module') before plain statement

````rust
    /// The file lives under the workspace root but no crate in the
    /// toolchain's graph reaches it, so every semantic query about it
    /// silently resolves to nothing
    ///
    /// This covers both a file the toolchain interned without attaching to
    /// a module and a file under the root the toolchain never interned at
    /// all, such as one a build script generated after the load.
````

</details>

File: crates/whisker-types/src/coverage_gap.rs [hunk 8]
```diff
@@ -0,0 +1,222 @@
...
+    /// toolchain's crate graph reaches it
+    ///
+    /// This also covers a file the toolchain never loaded.
+    Unreachable {
+        /// The workspace root the provider loaded
+        root: Arc<Path>,
+    },
+
+    /// The toolchain was configured to exclude the file
```
martijnarts, line 27: Rewritten. Rules broken:

- Implementation detail ('holds the file's identity but deliberately not its contents') without saying the plain WHAT first

````rust
    /// The toolchain holds the file's identity but deliberately not its
    /// contents, because it was configured to exclude it
````

</details>

**Review by martijnarts — COMMENTED** [review]

Not shown inline: `crates/whisker-types/src/decorated_tree.rs` fn DecoratedTree::decorations_mut — those blocks have no added line to anchor to (deleted content, or text a lower branch owns).

File: crates/whisker-types/src/coverage_gap.rs [hunk 9]
```diff
@@ -0,0 +1,222 @@
...
+        root: Arc<Path>,
+    },
+
+    /// The text the pipeline parsed differs from the text the toolchain
+    /// analyzed, so byte offsets from one do not address the other
+    StaleSource,
+}
+
+impl CoverageGap {
+    /// Returns the remedy to suggest to the user for this gap
```
martijnarts, line 39: Rewritten. Rules broken:

- 33-word sentence ('The remedy is separate from Display because...')

````rust
    /// Returns the remedy to suggest to the user for this gap
    ///
    /// The remedy is separate from [`Display`] because a caller that
    /// reports many gaps at once wants every reason but only the distinct
    /// remedies, which it can collect by de-duplicating these strings.
    ///
    /// # Examples
    ///
    /// ```
    /// use whisker_types::CoverageGap;
    ///
    /// assert_eq!(CoverageGap::StaleSource.help(), "re-run whisker");
    /// ```
    ///
    /// [`Display`]: std::fmt::Display
````

</details>

File: crates/whisker-types/src/decorated_tree.rs [hunk 10]
```diff
@@ -35,6 +35,13 @@ impl DecoratedTree {
...
 
+    /// Merges provider decorations into this tree
```
martijnarts, line 38: Rewritten. Rules broken:

- At this branch DecorationProvider receives the tree mutably and no pipeline calls merge_decorations, so 'the only path available to a DecorationProvider' is false
- 30-word sentence

````rust
    /// Merges provider decorations into this tree
    ///
    /// This is the only path available to a [`DecorationProvider`], which
    /// receives the tree immutably, and the pipeline calls it only after
    /// establishing that some provider claimed the file.
    ///
    /// [`DecorationProvider`]: crate::DecorationProvider
````

</details>

File: crates/whisker-types/src/decoration_map.rs [hunk 11]
```diff
@@ -47,6 +47,17 @@ impl DecorationMap {
             .and_then(|d| d.value.downcast_ref())
     }
 
+    /// Merges another map into this one
```
martijnarts, line 50: Rewritten. Rules broken:

- 38-word sentence ('Entries already present win: ... the provider consulted first is the one lint rules see')
- Explains pipeline consequences unrelated to this method's own behavior

````rust
    /// Merges another map into this one
    ///
    /// Entries already present win: [`DecorationMap::get`] returns the
    /// first decoration of a type for a node, so when two providers
    /// decorate the same node with the same type, the provider consulted
    /// first is the one lint rules see. Both remain reachable through
    /// [`DecorationMap::get_all`].
````

</details>

File: crates/whisker-types/src/provider_name.rs [hunk 12]
```diff
@@ -0,0 +1,78 @@
+/// Identifies a decoration provider in user-facing diagnostics
```
martijnarts, line 1: Rewritten. Rules broken:

- 27-word sentence
- Wordy rationale for a simple newtype ('the same shape RuleId uses for the same reason')

````rust
/// Identifies a decoration provider in user-facing diagnostics
///
/// Providers are linked in at compile time, not discovered at runtime, so
/// the name is always static: the same shape [`RuleId`] uses for the same
/// reason.
///
/// # Examples
///
/// ```
/// use whisker_types::ProviderName;
///
/// assert_eq!(ProviderName("rust").to_string(), "rust");
/// ```
///
/// [`RuleId`]: crate::RuleId
````

</details>

File: crates/whisker-types/src/uncovered_file.rs [hunk 13]
```diff
@@ -0,0 +1,200 @@
+use std::path::Path;
+use std::sync::Arc;
+
+use crate::{CoverageGap, ProviderName};
+
+/// Error returned when no decoration provider claims a file
```
martijnarts, line 6: Rewritten. Rules broken:

- 26-word sentence with omitted subject ('Carried through...')
- Essay ('The alternative, collapsing ... is the failure this type exists to prevent')
- Figure of speech (quoted inner voice 'I could not analyze this')

````rust
/// Error raised when no decoration provider claimed a file
///
/// Carried through [`anyhow::Error`] so a caller can recognize an analysis
/// failure by downcast and say something more useful about it than the
/// generic error path would. The alternative, collapsing "I could not
/// analyze this" into the same shape as "I analyzed this and found
/// nothing", is the failure this type exists to prevent.
///
/// # Examples
///
/// ```
/// use std::path::{Path, PathBuf};
/// use std::sync::Arc;
///
/// use whisker_types::{CoverageGap, ProviderName, UncoveredFile};
///
/// let root: Arc<Path> = Arc::from(PathBuf::from("/ws"));
/// let error = UncoveredFile::new(
///     PathBuf::from("stray.rs"),
///     vec![(ProviderName("rust"), CoverageGap::Unreachable { root })],
/// );
///
/// assert_eq!(
///     error.to_string(),
///     "no decoration provider covers this file, so semantic rules cannot run\n  \
///      rust: the file is inside the workspace at /ws but no crate in that workspace reaches it",
/// );
/// ```
///
/// [`anyhow::Error`]: anyhow::Error
````

</details>

File: crates/whisker-types/src/uncovered_file.rs [hunk 14]
```diff
@@ -0,0 +1,200 @@
...
+///
+/// [`anyhow::Error`]: anyhow::Error
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub struct UncoveredFile {
+    file: Arc<Path>,
+    gaps: Vec<(ProviderName, CoverageGap)>,
+}
+
+impl UncoveredFile {
+    /// Records that no provider claimed `file`, with each provider's reason
```
martijnarts, line 40: Rewritten. Rules broken:

- Filler ('at all')

````rust
    /// Records that no provider claimed `file`, with each provider's reason
    ///
    /// An empty `gaps` means no providers were consulted at all, which
    /// renders as its own message, not as an empty list.
````

</details>

File: crates/whisker-types/src/uncovered_file.rs [hunk 15]
```diff
@@ -0,0 +1,200 @@
...
+            gaps,
+        }
+    }
+
+    /// Returns the file that could not be analyzed
+    pub fn file(&self) -> &Path {
+        &self.file
+    }
+
+    /// Returns each provider consulted and its reason for declining
```
martijnarts, line 56: Rewritten. Rules broken:

- Omitted subject and verb ('Empty when...')
- Filler ('at all')

````rust
    /// Returns each provider consulted and its reason for declining
    ///
    /// Empty when no decoration providers were configured at all.
````

</details>

File: crates/whisker-types/src/coverage_gap.rs (hunk 1 above)
Comment (martijnarts) [reply crates/whisker-types/src/coverage_gap.rs]: You're right. The variant is now `OutsideRoot`, which matches the `root` field
it carries. The name says what happened: the provider loaded a root, and the
file lies outside it.

I took out the rest of the Cargo vocabulary in the file too, because a partial
rename leaves the crate just as Rust-specific. `Unreachable` said "no crate in
the toolchain's crate graph reaches it" and now says "nothing the toolchain
loaded reaches it". The three `root` field docs no longer say "workspace root".
I also dropped one sentence from `Unreachable`: "This also covers a file the
toolchain never loaded" became a tautology under the new wording.

`StaleSource` needed more than a word swap. It rendered "the file changed after
the workspace was loaded", which asserts an ordering the variant does not
establish. The variant is a mismatch: the text whisker parsed differs from the
text the toolchain analyzed. It now renders that.

The two remedies changed as well. `OutsideRoot` now suggests "check the file
from inside its own project". `Unreachable` suggests "reference the file from a
source the toolchain already loads", where it used to name a crate's module
tree. Say the word if you want the concrete Rust wording back in those two
strings. They are the one place the Cargo terms paid off. `help()` returns a
single static string per variant, so a remedy cannot vary by provider today.

Nothing outside `whisker-types` used the variant, so the rename touched
`coverage_gap.rs` and `uncovered_file.rs` only. I left
`[workspace.metadata.whisker]`, `WhiskerConfig`, and the rust-analyzer workspace
load alone, since those are Cargo-specific on purpose.

File: crates/whisker-types/src/decorated_tree.rs (hunk 2 above)
Comment (martijnarts) [reply crates/whisker-types/src/decorated_tree.rs]: `#[cfg(test)]` does not work here for three separate reasons.

It applies only to the crate under test, so `whisker-testing` cannot see the
method, and `whisker_testing::decorate` calls `decorations_mut` from its own
non-test code. The integration test in `whisker-macros/tests/decoration.rs` is a
separate crate too, so it cannot see the method either. On this branch the
method also has a production caller: `whisker-rust/src/provider.rs` inserts
every resolved decoration through `tree.decorations_mut()`.

So I left the method public and changed nothing. A cargo feature that
`whisker-testing` turns on answers the first two reasons, but not the third. It
only pays off once providers return decorations by value. The doc already names
the audience: "for providers to populate".

File: crates/whisker-types/src/provider_name.rs (hunk 3 above)
Comment (martijnarts) [reply crates/whisker-types/src/provider_name.rs:16]: Filed as https://github.com/aonyx-ai/whisker/issues/232. `ProviderName` keeps its public field in this PR.

The issue proposes a `lint.pub-field` rule. It flags public fields in
named-field structs and in tuple structs, and it points the author at a `getset`
derive. It records your wider framing and lists the two cases we need to settle
first: enum variant fields and `RuleId`.

One thing surfaced while writing it: `bool_param` misses tuple-struct fields
today, because it only matches `field_declaration` nodes. The issue notes that
too.

File: crates/whisker-types/src/decorated_tree.rs [hunk 16]
```diff
@@ -29,12 +29,30 @@ impl DecoratedTree {
...
-    /// populate
+    /// Returns a mutable reference to the decoration map for test helpers
+    /// that hand-build a decorated tree
+    ///
+    /// A [`DecorationProvider`] cannot reach this: it is handed the tree
+    /// immutably and returns its decorations instead, which the pipeline
+    /// applies with [`DecoratedTree::merge_decorations`]. This escape hatch
+    /// exists so a test can stage a decoration without loading a toolchain.
+    ///
+    /// [`DecorationProvider`]: crate::DecorationProvider
```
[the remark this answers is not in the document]
martijnarts: You were right. This comment sits against `7235dbd`, where the doc read:

```rust
/// Returns a mutable reference to the decoration map for test helpers
/// that hand-build a decorated tree
///
/// A [`DecorationProvider`] cannot reach this: it is handed the tree
/// immutably and returns its decorations instead, which the pipeline
/// applies with [`DecoratedTree::merge_decorations`]. This escape hatch
/// exists so a test can stage a decoration without loading a toolchain.
```

That described a contract this branch does not have. `decorate` takes
`&mut DecoratedTree` here, and `crates/whisker-rust/src/provider.rs:166-168`
writes through `decorations_mut`. The by-value contract lands in #213.

I rewrote the block in a later documentation pass. It now reads "Returns a
mutable reference to the decoration map for providers to populate", which is
true on this branch.

crates/whisker-types/src/coverage.rs
```diff
@@ -0,0 +1,45 @@
+use crate::{CoverageGap, DecorationMap};
+
+/// The result of offering one file to one decoration provider
+///
+/// Decorations travel back by value instead of being written through a
+/// `&mut DecoratedTree`. That is what makes a declining provider harmless:
+/// it never holds a handle it could mutate, so "declined" and "left partial
+/// decorations behind" cannot both be true.
+///
```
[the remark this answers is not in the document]
martijnarts: You were right. This comment sits against `7235dbd`, where the type doc's
second paragraph read:

```rust
/// Decorations travel back by value instead of being written through a
/// `&mut DecoratedTree`. That is what makes a declining provider harmless:
/// it never holds a handle it could mutate, so "declined" and "left partial
/// decorations behind" cannot both be true.
```

That is the #213 contract, not this branch's, so the doc described behavior
the code did not have.

I rewrote it in the same pass. The text now describes the type rather than
the call: a verdict from one provider for one file, and `NotCovered` carries
the reason. `coverage.rs` is byte-identical on this branch and on #213, so
the text survives the contract change.

I left the other half of your suggestion alone. The by-value flow is where
the type is headed, not what it is today. I want the doc to say what it is.

PR #213, opened 2026-08-05
Make a decoration provider state its coverage

martijnarts:
`DecorationProvider::decorate` returned `anyhow::Result<()>`, so `Ok(())` meant
both "I decorated this file" and "I have nothing for this file". The pipeline
read the first meaning, walked a completely undecorated tree, and the three
rules that depend on semantic decorations found nothing. Whisker reported the
file clean because it had been unable to analyze it, which the user cannot
tell apart from a clean bill of health.

Providers now return `Coverage`, and one that declines hands back the gap
explaining why. `Pipeline::run_on_source` refuses to walk a tree no provider
claimed. Decorations come back by value instead of being written through a
mutable handle, which makes "declined after writing half its decorations"
unrepresentable.

Answering honestly turned out to need three fixes to what the provider knows.
VFS membership was the wrong predicate: `parse_guess_edition` papers over a
missing module by guessing the current edition and resolving nothing, so a
file the toolchain interned but no crate reaches passed the guard and produced
nothing. Cargo leaves `test` out of a crate's cfg options, so every
`#[cfg(test)] mod tests` block resolved to nothing while whisker walked it and
reported it clean; those blocks hold roughly half the lines of this codebase.
And a file whose bytes are not UTF-8 is interned without its text, so
rust-analyzer panicked when something later read it, from inside the crate
around it, never the file whisker was asked about.

The CLI still reports a declined file as an ordinary per-file failure. What it
should say about one, and what happens to the diagnostics already collected,
comes next in the stack.

### DIFF
--- a/crates/whisker-core/src/pipeline.rs
+++ b/crates/whisker-core/src/pipeline.rs
@@ -1,7 +1,9 @@
 use std::path::Path;
 
 use anyhow::Context as _;
-use whisker_types::{DecoratedTree, DecorationProvider, Diagnostic, Language, LintPass};
+use whisker_types::{
+    Coverage, DecoratedTree, DecorationProvider, Diagnostic, Language, LintPass, UncoveredFile,
+};
 
 use crate::tree_walker;
 
@@ -32,7 +34,8 @@ impl Pipeline {
     ///
     /// # Errors
     ///
-    /// Returns an error if the file cannot be read, parsed, or decorated.
+    /// Returns an error if the file cannot be read. Any other error comes
+    /// from [`Pipeline::run_on_source`].
     pub fn run(
         &mut self,
         path: &Path,
@@ -47,9 +50,18 @@ impl Pipeline {
 
     /// Runs the pipeline on source text directly
     ///
+    /// The pipeline offers the file to every provider, and one claim is
+    /// enough. If no provider claims the file, the pipeline returns an
+    /// [`UncoveredFile`] error instead of a false clean result.
+    ///
+    /// Decorations merge in provider order. The first decoration of a
+    /// type on a node is the one lint rules see.
+    ///
     /// # Errors
     ///
-    /// Returns an error if parsing or decoration fails.
+    /// Returns an error if parsing fails, if a provider's toolchain
+    /// malfunctions, or if no provider claims the file. An empty provider
+    /// list counts as no claims.
     pub fn run_on_source(
         &mut self,
         source: &str,
@@ -64,10 +76,26 @@ impl Pipeline {
 
         let mut decorated = DecoratedTree::new(tree, source.to_string(), path.to_path_buf());
 
+        let mut covered = Vec::new();
+        let mut gaps = Vec::new();
+
         for provider in providers {
-            provider
-                .decorate(&mut decorated)
-                .context("decoration provider failed")?;
+            let coverage = provider
+                .decorate(&decorated)
+                .with_context(|| format!("run the `{}` decoration provider", provider.name()))?;
+
+            match coverage {
+                Coverage::Covered(decorations) => covered.push(decorations),
+                Coverage::NotCovered(gap) => gaps.push((provider.name(), gap)),
+            }
+        }
+
+        if covered.is_empty() {
+            return Err(UncoveredFile::new(path.to_path_buf(), gaps).into());
+        }
+
+        for decorations in covered {
+            decorated.merge_decorations(decorations);
         }
 
         Ok(tree_walker::walk(&decorated, passes))
@@ -93,9 +121,79 @@ pub fn detect_language(path: &Path) -> anyhow::Result<Language> {
 mod tests {
     use std::path::PathBuf;
 
-    use whisker_types::{DecoratedNode, Diagnostic, Language, RuleId, Severity};
+    use whisker_types::{
+        CoverageGap, DecoratedNode, DecorationMap, Diagnostic, Language, ProviderName, RuleId,
+        Severity,
+    };
 
     use super::*;
+
+    /// Claims every file and decorates nothing
+    struct Covering;
+
+    impl DecorationProvider for Covering {
+        fn name(&self) -> ProviderName {
+            ProviderName("covering")
+        }
+
+        fn decorate(&self, _tree: &DecoratedTree) -> anyhow::Result<Coverage> {
+            Ok(Coverage::Covered(DecorationMap::new()))
+        }
+    }
+
+    /// Claims nothing, so the pipeline must treat the file as unanalyzable
+    struct Declining;
+
+    impl DecorationProvider for Declining {
+        fn name(&self) -> ProviderName {
+            ProviderName("declining")
+        }
+
+        fn decorate(&self, _tree: &DecoratedTree) -> anyhow::Result<Coverage> {
+            Ok(Coverage::NotCovered(CoverageGap::StaleSource))
+        }
+    }
+
+    /// Claims every file and puts a known marker on the root node
+    struct Decorating(&'static str);
+
+    impl DecorationProvider for Decorating {
+        fn name(&self) -> ProviderName {
+            ProviderName("decorating")
+        }
+
+        fn decorate(&self, tree: &DecoratedTree) -> anyhow::Result<Coverage> {
+            let mut decorations = DecorationMap::new();
+            decorations.insert(tree.root_node().id(), Marker(self.0));
+            Ok(Coverage::Covered(decorations))
+        }
+    }
+
+    #[derive(Debug)]
+    struct Marker(&'static str);
+
+    /// Reports whichever [`Marker`] a provider left on the root node
+    ///
+    /// The assertion runs through a lint pass, so the test observes what
+    /// a real rule would see, not what the pipeline happened to store.
+    struct ReportMarker;
+
+    impl LintPass for ReportMarker {
+        fn check_node(&mut self, node: &DecoratedNode<'_>) -> Vec<Diagnostic> {
+            if node.kind() != "source_file" {
+                return Vec::new();
+            }
+            let Some(marker) = node.decoration::<Marker>() else {
+                return Vec::new();
+            };
+            vec![Diagnostic::new(
+                RuleId("test.marker"),
+                Severity::Warn,
+                marker.0.into(),
+                node.span(),
+            )]
+        }
+    }
 
     #[test]
     fn trait_send() {
@@ -134,12 +232,62 @@ mod tests {
     }
 
     #[test]
+    fn run_on_source_with_covering_and_declining_providers_returns_diagnostics() {
+        let ts_lang: tree_sitter::Language = tree_sitter_rust::LANGUAGE.into();
+        let mut pipeline = Pipeline::new(&ts_lang).unwrap();
+        let mut passes: Vec<Box<dyn LintPass>> = vec![Box::new(ReportMarker)];
+
+        let diagnostics = pipeline
+            .run_on_source(
+                "fn main() {}",
+                Path::new("test.rs"),
+                &[
+                    &Declining as &dyn DecorationProvider,
+                    &Decorating("covered"),
+                ],
+                &mut passes,
+            )
+            .expect("one covering provider should be enough");
+
+        assert_eq!(diagnostics.len(), 1);
+        assert_eq!(diagnostics[0].message(), "covered");
+    }
+
+    #[test]
+    fn run_on_source_with_declining_provider_returns_uncovered_file() {
+        let ts_lang: tree_sitter::Language = tree_sitter_rust::LANGUAGE.into();
+        let mut pipeline = Pipeline::new(&ts_lang).unwrap();
+
+        let error = pipeline
+            .run_on_source(
+                "fn main() {}",
+                Path::new("test.rs"),
+                &[&Declining as &dyn DecorationProvider],
+                &mut Vec::new(),
+            )
+            .expect_err("a declining provider should not cover the file");
+
+        let uncovered = error
+            .downcast_ref::<UncoveredFile>()
+            .expect("should be an UncoveredFile");
+        assert_eq!(uncovered.file(), Path::new("test.rs"));
+        assert_eq!(uncovered.gaps().len(), 1);
+        assert_eq!(uncovered.gaps()[0].0, ProviderName("declining"));
+        assert_eq!(uncovered.gaps()[0].1, CoverageGap::StaleSource);
+    }
+
+    #[test]
     fn run_on_source_with_empty_passes_produces_no_diagnostics() {
         let ts_lang: tree_sitter::Language = tree_sitter_rust::LANGUAGE.into();
         let mut pipeline = Pipeline::new(&ts_lang).unwrap();
 
         let diagnostics = pipeline
-            .run_on_source("fn main() {}", Path::new("test.rs"), &[], &mut Vec::new())
+            .run_on_source(
+                "fn main() {}",
+                Path::new("test.rs"),
+                &[&Covering as &dyn DecorationProvider],
+                &mut Vec::new(),
+            )
             .unwrap();
 
         assert!(diagnostics.is_empty());
@@ -168,11 +316,84 @@ mod tests {
         let mut passes: Vec<Box<dyn whisker_types::LintPass>> = vec![Box::new(AlwaysWarn)];
 
         let diagnostics = pipeline
-            .run_on_source("fn main() {}", Path::new("test.rs"), &[], &mut passes)
+            .run_on_source(
+                "fn main() {}",
+                Path::new("test.rs"),
+                &[&Covering as &dyn DecorationProvider],
+                &mut passes,
+            )
             .unwrap();
 
         assert_eq!(diagnostics.len(), 1);
         assert_eq!(diagnostics[0].message(), "found function");
+    }
+
+    #[test]
+    fn run_on_source_with_no_providers_returns_uncovered_file() {
+        let ts_lang: tree_sitter::Language = tree_sitter_rust::LANGUAGE.into();
+        let mut pipeline = Pipeline::new(&ts_lang).unwrap();
+
+        let error = pipeline
+            .run_on_source("fn main() {}", Path::new("test.rs"), &[], &mut Vec::new())
+            .expect_err("no providers means nothing can be analyzed");
+
+        let uncovered = error
+            .downcast_ref::<UncoveredFile>()
+            .expect("should be an UncoveredFile");
+        assert!(uncovered.gaps().is_empty());
+        assert!(
+            uncovered
+                .to_string()
+                .contains("no decoration providers were configured"),
+            "unexpected message: {uncovered}"
+        );
+    }
+
+    #[test]
+    fn run_on_source_with_two_decorating_providers_keeps_first_decoration() {
+        let ts_lang: tree_sitter::Language = tree_sitter_rust::LANGUAGE.into();
+        let mut pipeline = Pipeline::new(&ts_lang).unwrap();
+        let mut passes: Vec<Box<dyn LintPass>> = vec![Box::new(ReportMarker)];
+
+        let diagnostics = pipeline
+            .run_on_source(
+                "fn main() {}",
+                Path::new("test.rs"),
+                &[
+                    &Decorating("first") as &dyn DecorationProvider,
+                    &Decorating("second"),
+                ],
+                &mut passes,
+            )
+            .expect("both providers cover the file");
+
+        assert_eq!(diagnostics.len(), 1);
+        assert_eq!(diagnostics[0].message(), "first");
+    }
+
+    #[test]
+    fn run_on_source_with_uncovered_file_does_not_run_passes() {
+        struct Exploding;
+        impl whisker_types::LintPass for Exploding {
+            fn check_node(&mut self, _node: &DecoratedNode<'_>) -> Vec<Diagnostic> {
+                panic!("lint passes must not run on an uncovered file");
+            }
+        }
+
+        let ts_lang: tree_sitter::Language = tree_sitter_rust::LANGUAGE.into();
+        let mut pipeline = Pipeline::new(&ts_lang).unwrap();
+        let mut passes: Vec<Box<dyn whisker_types::LintPass>> = vec![Box::new(Exploding)];
+
+        let error = pipeline
+            .run_on_source(
+                "fn main() {}",
+                Path::new("test.rs"),
+                &[&Declining as &dyn DecorationProvider],
+                &mut passes,
+            )
+            .expect_err("a declining provider should not cover the file");
+
+        assert!(error.downcast_ref::<UncoveredFile>().is_some());
     }
 
     mod prop {
@@ -206,7 +427,7 @@ mod tests {
                 let result = pipeline.run_on_source(
                     &source,
                     Path::new("test.rs"),
-                    &[],
+                    &[&Covering as &dyn DecorationProvider],
                     &mut Vec::new(),
                 );
 
@@ -224,11 +445,29 @@ mod tests {
                 let result = pipeline.run_on_source(
                     &source,
                     Path::new("test.rs"),
-                    &[],
+                    &[&Covering as &dyn DecorationProvider],
                     &mut Vec::new(),
                 );
 
                 prop_assert!(result.is_ok());
+            }
+
+            #[test]
+            fn run_on_source_without_providers_always_reports_uncovered(
+                source in "(fn [a-z]+\\(\\) \\{\\}\n){0,5}",
+            ) {
+                let ts_lang: tree_sitter::Language = tree_sitter_rust::LANGUAGE.into();
+                let mut pipeline = Pipeline::new(&ts_lang).unwrap();
+
+                let result = pipeline.run_on_source(
+                    &source,
+                    Path::new("test.rs"),
+                    &[],
+                    &mut Vec::new(),
+                );
+
+                let error = result.expect_err("no providers can never cover a file");
+                prop_assert!(error.downcast_ref::<UncoveredFile>().is_some());
             }
         }
     }
--- a/crates/whisker-rust/src/provider.rs
+++ b/crates/whisker-rust/src/provider.rs
@@ -1,15 +1,19 @@
 use std::path::Path;
-use std::sync::Mutex;
+use std::sync::{Arc, Mutex};
 
 use anyhow::Context as _;
 use ra_ap_hir::{Adt, HasAttrs, HirDisplay, Semantics, attach_db};
-use ra_ap_ide_db::RootDatabase;
+use ra_ap_ide_db::base_db::SourceDatabase as _;
+use ra_ap_ide_db::{ChangeWithProcMacros, RootDatabase};
 use ra_ap_load_cargo::{LoadCargoConfig, ProcMacroServerChoice};
-use ra_ap_project_model::CargoConfig;
+use ra_ap_project_model::{CargoConfig, ProjectManifest, ProjectWorkspace};
 use ra_ap_syntax::AstNode;
 use ra_ap_syntax::ast;
-use ra_ap_vfs::Vfs;
-use whisker_types::{DecoratedNode, DecoratedTree, DecorationProvider};
+use ra_ap_vfs::{AbsPathBuf, FileExcluded, FileId, Vfs, VfsPath};
+use whisker_types::{
+    Coverage, CoverageGap, DecoratedNode, DecoratedTree, DecorationMap, DecorationProvider,
+    ProviderName,
+};
 
 use crate::decorations::{AdtFlags, FnSignature, ResolvedType};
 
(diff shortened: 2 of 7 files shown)


martijnarts:
Every doc comment this PR adds was re-checked against three rule sets and rewritten where it failed. 19 blocks changed here.

Violations found here: 16 reviewer verdicts from #208, 26 ASD-STE100, 15 AI-writing tells.

|---|---|---|
| `crates/whisker-core/src/pipeline.rs` | Pipeline::run | Em-dash aside ("file — see ..."); 25-word sentence at the cap |
| `crates/whisker-core/src/pipeline.rs` | Pipeline::run_on_source | 33-word sentence in paragraph 1; -ing clauses ("instead of walking", "demanding that"); Multi-workspace paragraph is design rationale, basically unrelated to what the function does; Passi... |
| `crates/whisker-core/src/pipeline.rs` | test helper struct ReportMarker | -ing subject clause ("Routing the assertion ... keeps"); Figurative phrase ("keeps the test honest") |
| `crates/whisker-rust/src/provider.rs` | RustDecorationProvider::empty | 32-word and 40-word sentences; Essay about "the honest description of what an unloaded provider can do"; -ing summary clause ("provider holding"); The Examples match is oversized for a co... |
| `crates/whisker-rust/src/provider.rs` | RustDecorationProvider::gap_for_unknown_file | 30-word and 33-word sentences; Figurative phrase ("sends the user looking for a second workspace"); Says the rationale before the behavior; what the function returns comes last |
| `crates/whisker-rust/src/provider.rs` | RustDecorationProvider::load | Implementation-detaily essay about load_workspace_at internals; Set_test paragraph is an essay about codebase halves and editors, unrelated to the caller's contract; 33-word and 40-word s... |
| `crates/whisker-rust/src/provider.rs` | give_text_to_files_that_have_none | Idiom ("drops its text on the floor"); 45-word sentence with emphasis markup (*siblings*); -ing subjects ("Transcribing", "Covering the rest", "Repairing only Rust source"); "Extension is... |
| `crates/whisker-rust/src/provider.rs` | impl DecorationProvider for RustDecorationProvider — decorate | Very implementation detaily; the guard-ordering paragraphs document internals the caller never needs; 45-word and 52-word sentences; "Only the first two conditions are guards" essay is ba... |
| `crates/whisker-rust/src/provider.rs` | resolve_targets | 26-word sentence over the cap |
| `crates/whisker-rust/src/provider.rs` | struct RustDecorationProvider | Figurative phrase ("can honestly say anything about", "kept around"); 29-word sentence in the added paragraph |
| `crates/whisker-rust/src/provider.rs` | test decorate_with_excluded_file_reports_excluded_by_toolchain | Test doc must be one or two sentences; this is a two-paragraph essay; 30-word and 34-word sentences; Rhetorical flourish ("its own test instead of trust", "cannot decline to have an opini... |
| `crates/whisker-rust/tests/fixtures/sample_project/orphan.rs` | module doc (orphan.rs) | Em-dash aside pair ("— known to the toolchain, reachable by no crate —"); -ing clause ("while resolving nothing about it") |
| `crates/whisker-rust/tests/provider_integration.rs` | fn not_utf8_project | Idiom ("on the fly", "take the whole test binary down"); 41-word and 40-word sentences; Passive with known agent ("is generated", "are reached") |
| `crates/whisker-rust/tests/provider_integration.rs` | test decorate_with_a_module_that_is_not_utf8_covers_its_siblings | Test doc too long; the mechanism fits in two sentences; 52-word sentence; Idiom ("take the rest of its crate down with it", "turns this red") |
| `crates/whisker-rust/tests/provider_integration.rs` | test decorate_with_code_under_cfg_test_resolves_types | 28-word sentence; "the difference between linting a project and linting the part that is not tested" is a rhetorical essay; Significance inflation ("roughly half the lines of a Rust codeb... |
| `crates/whisker-rust/tests/provider_integration.rs` | test decorate_with_generated_file_under_root_reports_unreachable | 33-word sentence; Test doc too long; keep the pin and one mechanism; Rhetorical flourish ("a remedy the user cannot act on") |
| `crates/whisker-types/src/decoration_provider.rs` | DecorationProvider::decorate | "borrowed immutably, returned by value" documents the signature, no one thought otherwise; 30-word sentence; Em-dash aside in # Errors; -ing subject ("returning an error for it would abor... |
| `crates/whisker-types/src/decoration_provider.rs` | DecorationProvider::name | Passive voice with a known agent ("is identified by in diagnostics") |
| `crates/whisker-types/src/decoration_provider.rs` | trait DecorationProvider | 34-word sentence in paragraph 1; 46-word semicolon sentence in paragraph 2; Paragraph 2 is an essay; the point fits in two short sentences; Significance inflation ("not optional and not i... |

crates/whisker-core/src/pipeline.rs
```diff
@@ -32,7 +34,8 @@ impl Pipeline {
     ///
     /// # Errors
     ///
-    /// Returns an error if the file cannot be read, parsed, or decorated.
+    /// Returns an error if the file cannot be read. Any other error comes
```
martijnarts, line 37: Rewritten. Rules broken:

- Em-dash aside ("file — see ...")
- 25-word sentence at the cap

````rust
    /// Runs the full pipeline on a source file
    ///
    /// # Errors
    ///
    /// Returns an error if the file cannot be read or parsed, if a
    /// provider's toolchain malfunctions, or if no provider claims the
    /// file — see [`Pipeline::run_on_source`].
````

</details>

File: crates/whisker-core/src/pipeline.rs [hunk 2]
```diff
@@ -47,9 +50,18 @@ impl Pipeline {
 
     /// Runs the pipeline on source text directly
     ///
+    /// The pipeline offers the file to every provider, and one claim is
```
martijnarts, line 53: Rewritten. Rules broken:

- 33-word sentence in paragraph 1
- -ing clauses ("instead of walking", "demanding that")
- Multi-workspace paragraph is design rationale, basically unrelated to what the function does
- Passive voice ("is offered", "are merged")

````rust
    /// Runs the pipeline on source text directly
    ///
    /// Every provider is offered the file. If none claims it, the pipeline
    /// returns an [`UncoveredFile`] error instead of walking a tree with
    /// no semantic information, which would report the file clean while
    /// the semantic rules silently did nothing.
    ///
    /// One claim is enough. Two providers loaded for two workspaces can
    /// never both claim the same file, so demanding that all of them do
    /// would make the multi-workspace case impossible.
    ///
    /// Providers are merged in the order they are given, and the first
    /// decoration of a type on a node is the one lint rules see.
    ///
    /// # Errors
    ///
    /// Returns an error if parsing fails, if a provider's toolchain
    /// malfunctions, or if no provider claims the file, including the
    /// case where no providers were configured at all.
````

</details>

File: crates/whisker-core/src/pipeline.rs [hunk 3]
```diff
@@ -93,10 +121,80 @@ pub fn detect_language(path: &Path) -> anyhow::Result<Language> {
...
+            let mut decorations = DecorationMap::new();
+            decorations.insert(tree.root_node().id(), Marker(self.0));
+            Ok(Coverage::Covered(decorations))
+        }
+    }
+
+    #[derive(Debug)]
+    struct Marker(&'static str);
+
+    /// Reports whichever [`Marker`] a provider left on the root node
```
martijnarts, line 175: Rewritten. Rules broken:

- -ing subject clause ("Routing the assertion ... keeps")
- Figurative phrase ("keeps the test honest")

````rust
    /// Reports whichever [`Marker`] a provider left on the root node
    ///
    /// Routing the assertion through a lint pass keeps the test honest: it
    /// observes what a real rule would see, not what the pipeline happened
    /// to store.
````

</details>

**Review by martijnarts — COMMENTED** [review]

File: crates/whisker-types/src/decoration_provider.rs [hunk 4]
```diff
@@ -1,32 +1,46 @@
-use crate::DecoratedTree;
+use crate::{Coverage, DecoratedTree, ProviderName};
 
 /// Attaches semantic decorations to a parsed syntax tree
 ///
 /// Decoration providers bridge between a language toolchain (e.g.
 /// rust-analyzer) and the platform's decoration system. A provider reads
-/// the syntax tree, queries the toolchain for semantic information, and
-/// inserts decorations into the tree's [`DecorationMap`].
+/// the syntax tree, queries the toolchain, and returns decorations
```
martijnarts, line 7: Rewritten. Rules broken:

- 34-word sentence in paragraph 1
- 46-word semicolon sentence in paragraph 2
- Paragraph 2 is an essay; the point fits in two short sentences
- Significance inflation ("not optional and not inferable")

````rust
/// Attaches semantic decorations to a parsed syntax tree
///
/// Decoration providers bridge between a language toolchain (e.g.
/// rust-analyzer) and the platform's decoration system. A provider reads
/// the syntax tree, queries the toolchain for semantic information, and
/// returns the decorations it produced together with a verdict on whether
/// it had anything to say about the file at all.
///
/// The verdict is not optional and not inferable from the decorations. A
/// provider that returned no decorations because the file has nothing to
/// decorate, and a provider that returned no decorations because it has
/// never heard of the file, are indistinguishable by their output and must
/// be distinguishable by their verdict; otherwise whisker reports a file
/// clean that it could not read the meaning of.
````

</details>

File: crates/whisker-types/src/decoration_provider.rs [hunk 5]
```diff
@@ -1,32 +1,46 @@
...
+/// together with a [`Coverage`] verdict.
 ///
-/// [`DecorationMap`]: crate::DecorationMap
+/// The verdict tells the caller whether the provider could analyze the
+/// file. An empty decoration set is ambiguous: a file with nothing to
+/// decorate and an unknown file produce the same output.
 pub trait DecorationProvider: Send + Sync {
-    /// Populates decorations on the given tree
+    /// Returns the name that identifies this provider in diagnostics
```
martijnarts, line 14: Rewritten. Rules broken:

- Passive voice with a known agent ("is identified by in diagnostics")

````rust
    /// Returns the name this provider is identified by in diagnostics
````

</details>

File: crates/whisker-types/src/decoration_provider.rs [hunk 6]
```diff
@@ -1,32 +1,46 @@
...
+    fn name(&self) -> ProviderName;
+
+    /// Decorates the file, or explains why this provider cannot
```
martijnarts: Rewritten. Rules broken:

- "borrowed immutably, returned by value" documents the signature, no one thought otherwise
- 30-word sentence
- Em-dash aside in # Errors
- -ing subject ("returning an error for it would abort")
- Near-duplicate of the impl's # Errors sentence in provider.rs

````rust
    /// Decorates the file, or explains why this provider cannot
    ///
    /// The tree is borrowed immutably and decorations are returned by
    /// value, so a provider that declines has no way to leave partial
    /// state behind and the caller can discard its work without
    /// inspection.
    ///
    /// # Errors
    ///
    /// Returns an error only when the toolchain itself malfunctions — a
    /// poisoned lock, a path that cannot be made absolute. A file the
    /// provider simply has no information about is
    /// [`Coverage::NotCovered`], not an error: returning an error for it
    /// would abort a whole run on the first stray file.
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 7]
```diff
@@ -20,30 +24,46 @@ use crate::decorations::{AdtFlags, FnSignature, ResolvedType};
 /// decorate phase. The workspace load is expensive and happens once; the
 /// per-file decoration pass is relatively cheap.
 ///
+/// The provider stores the workspace root. Every gap except
```
martijnarts, line 27: Rewritten. Rules broken:

- Figurative phrase ("can honestly say anything about", "kept around")
- 29-word sentence in the added paragraph

````rust
/// Decoration provider for Rust using rust-analyzer
///
/// Loads a Cargo workspace into a rust-analyzer database at construction
/// time, then attaches type information to tree-sitter nodes during the
/// decorate phase. The workspace load is expensive and happens once; the
/// per-file decoration pass is relatively cheap.
///
/// The loaded workspace is also the boundary of what this provider can
/// honestly say anything about, so the root is kept around: every
/// [`CoverageGap`] it reports names the workspace that declined.
///
/// # Examples
///
/// ```ignore
/// let provider = RustDecorationProvider::load(Path::new("."))?;
/// let coverage = provider.decorate(&tree)?;
/// ```
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 8]
```diff
@@ -20,30 +24,46 @@ use crate::decorations::{AdtFlags, FnSignature, ResolvedType};
...
+    root: Arc<Path>,
 }
 
 impl RustDecorationProvider {
+    /// The name this provider is reported under in diagnostics
+    const NAME: ProviderName = ProviderName("rust");
+
     /// Loads a Cargo workspace for semantic analysis
     ///
+    /// rust-analyzer discovers the nearest manifest above
```
martijnarts, line 48: Rewritten. Rules broken:

- Implementation-detaily essay about load_workspace_at internals
- Set_test paragraph is an essay about codebase halves and editors, unrelated to the caller's contract
- 33-word and 40-word sentences
- Rhetorical filler ("helps nobody", "not necessarily the answer")
- "Each stage names itself, because their remedies differ" documents error-message design, no one asked

````rust
    /// Loads a Cargo workspace for semantic analysis
    ///
    /// `workspace_root` is a starting point, not necessarily the answer:
    /// rust-analyzer discovers the nearest manifest above it and Cargo then
    /// expands that to the whole workspace. This spells out the steps that
    /// [`ra_ap_load_cargo::load_workspace_at`] would otherwise perform in
    /// one call, because the workspace it settles on is what every
    /// [`CoverageGap`] this provider reports has to name. Telling a user
    /// their file is outside "the workspace at src/lib.rs" helps nobody.
    ///
    /// This is an expensive operation that builds rust-analyzer's internal
    /// database from the project's Cargo.toml. Call this once at startup.
    ///
    /// `set_test` is on, which puts `test` in the cfg options of every crate
    /// the workspace owns. Cargo's own default is off, so leaving it there
    /// would resolve a `#[cfg(test)] mod tests` block to nothing: whisker
    /// would still walk the file and the rules that need types would quietly
    /// find none, over the half of a typical Rust codebase that lives in
    /// those blocks. Editors running rust-analyzer turn it on for the same
    /// reason.
    ///
    /// # Errors
    ///
    /// Returns an error if the current directory cannot be read, if
    /// `workspace_root` is not valid UTF-8, if no Cargo project can be
    /// discovered from it, or if any stage of the load fails: resolving the
    /// workspace, running its build scripts, or building the analysis
    /// database. Each stage names itself, because their remedies differ.
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 9]
```diff
@@ -54,50 +74,170 @@ impl RustDecorationProvider {
...
 
         Ok(Self {
             db: Mutex::new(db),
             vfs,
+            root,
         })
     }
 
-    /// Creates a no-op provider that attaches no decorations
+    /// Creates a provider that holds an empty workspace
```
martijnarts, line 118: Rewritten. Rules broken:

- 32-word and 40-word sentences
- Essay about "the honest description of what an unloaded provider can do"
- -ing summary clause ("provider holding")
- The Examples match is oversized for a constructor

````rust
    /// Creates a provider holding an empty workspace
    ///
    /// Its VFS knows no files and its root is relative, so no absolute file
    /// path can ever lie under it: every file is declined with
    /// [`CoverageGap::OutsideWorkspace`]. That is the honest description of
    /// what an unloaded provider can do, and it is why this constructor
    /// cannot be used as a fallback when [`RustDecorationProvider::load`]
    /// fails: the pipeline would refuse every file, not report them all
    /// clean.
    ///
    /// # Examples
    ///
    /// ```ignore
    /// let provider = RustDecorationProvider::empty();
    ///
    /// match provider.decorate(&tree)? {
    ///     Coverage::Covered(_) => unreachable!("an empty VFS covers nothing"),
    ///     Coverage::NotCovered(CoverageGap::OutsideWorkspace { root }) => {
    ///         eprintln!("no workspace loaded at {}", root.display());
    ///     }
    ///     Coverage::NotCovered(gap) => eprintln!("declined: {gap}"),
    /// }
    /// ```
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 10]
```diff
@@ -54,50 +74,170 @@ impl RustDecorationProvider {
...
+    /// ```
     pub fn empty() -> Self {
         Self {
             db: Mutex::new(RootDatabase::default()),
             vfs: Vfs::default(),
+            root: Arc::from(Path::new(".")),
+        }
+    }
+
+    /// Classifies a file rust-analyzer never interned
```
martijnarts, line 139: Rewritten. Rules broken:

- 30-word and 33-word sentences
- Figurative phrase ("sends the user looking for a second workspace")
- Says the rationale before the behavior; what the function returns comes last

````rust
    /// Classifies a file rust-analyzer never interned
    ///
    /// A path under the loaded root is one the toolchain could have known
    /// about and did not: a file no crate reaches, or one that appeared
    /// after the load, such as build-script output under `target`. Calling
    /// that "outside the workspace" while printing a path plainly inside it
    /// sends the user looking for a second workspace that does not exist,
    /// so only a path that really is elsewhere gets that verdict.
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 11]
```diff
@@ -54,50 +74,170 @@ impl RustDecorationProvider {
...
+            }
+        } else {
+            CoverageGap::OutsideWorkspace {
+                root: Arc::clone(&self.root),
+            }
         }
     }
 }
 
+/// Records lossy UTF-8 text for interned files under `root` that have none
```
martijnarts, line 157: Rewritten. Rules broken:

- Idiom ("drops its text on the floor")
- 45-word sentence with emphasis markup (*siblings*)
- -ing subjects ("Transcribing", "Covering the rest", "Repairing only Rust source")
- "Extension is not one of the tests" buries the point; say what the function checks plainly

````rust
/// Gives a text to every interned file under `root` that has none
///
/// [`ra_ap_load_cargo`] interns a file whose bytes are not UTF-8 but drops
/// its text on the floor, and rust-analyzer's file store panics on a file
/// whose text was never recorded. The panic cannot be guarded where whisker
/// asks its questions: rust-analyzer reads a module's text while building
/// the definition map of the crate around it, so asking about a file is
/// enough to kill the process over one of its *siblings*, and what the user
/// sees is a file id with no path. Transcribing the bytes lossily leaves
/// the database total, which is the only shape in which the rest of the
/// workspace can be analyzed at all. The file itself stays undecoratable:
/// its lossy text cannot equal the text whisker parsed, so the source-match
/// guard in `decorate` declines it.
///
/// Only files under `root` are read. Covering the rest would mean re-reading
/// the sysroot and every dependency on every run, and a file in either that
/// rust-analyzer would parse is UTF-8 already, or the dependency would not
/// have compiled.
///
/// Extension is not one of the tests. Rust source is the only thing whisker
/// decorates, but it is not the only thing whose text rust-analyzer reads:
/// `include_str!` reads whatever path it is given, and the load interns
/// `.toml` and `.md` beside the `.rs` files. Repairing only Rust source
/// would leave `include_str!("notes.md")` able to end the run, and the few
/// manifests a workspace holds cost nothing to check.
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 12]
```diff
@@ -54,50 +74,170 @@ impl RustDecorationProvider {
...
+    db.apply_change(change);
+}
+
 impl DecorationProvider for RustDecorationProvider {
-    /// Attaches type decorations to the tree's nodes
+    fn name(&self) -> ProviderName {
+        Self::NAME
+    }
+
+    /// Attaches type decorations to the tree's nodes, or declines the file
```
martijnarts, line 208: Rewritten. Rules broken:

- Very implementation detaily; the guard-ordering paragraphs document internals the caller never needs
- 45-word and 52-word sentences
- "Only the first two conditions are guards" essay is basically unrelated to what decorate returns

````rust
    /// Attaches type decorations to the tree's nodes, or declines the file
    ///
    /// Four conditions must hold before a single decoration is produced,
    /// and their order matters. The file must be interned by the VFS, or
    /// rust-analyzer has never heard of it. It must not be excluded, or
    /// every question the database can be asked about it panics: the VFS
    /// keeps an excluded file's identity and neither its contents nor a
    /// source root, so the check has to come before the database is touched
    /// at all, not merely before its text is read. It must belong to a
    /// module, because `parse_guess_edition` papers over a missing module by
    /// guessing the current edition and then resolves nothing. And the text
    /// the pipeline parsed must equal the text the database holds, because
    /// the byte offsets carried by every target index into rust-analyzer's
    /// parse of that text.
    ///
    /// Only the first two conditions are guards. The module check protects
    /// nothing, because establishing the module reads whatever sibling
    /// modules the crate declares; a file whose text the load dropped would
    /// already have brought the process down by then, which is why
    /// [`give_text_to_files_that_have_none`] repairs the database at load
    /// time, so the read here needs no careful ordering.
    ///
    /// # Errors
    ///
    /// Returns an error if the file path cannot be made absolute or the
    /// database lock is poisoned. A file this provider knows nothing about
    /// is a [`Coverage::NotCovered`] verdict, not an error.
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 13]
```diff
@@ -106,77 +246,85 @@ impl DecorationProvider for RustDecorationProvider {
...
+
+        Ok(coverage)
     }
 }
 
-enum Decoration {
-    Type(usize, ResolvedType),
-    Adt(usize, AdtFlags),
-    FnSig(usize, FnSignature),
+/// Resolves every collected target against rust-analyzer's parse
```
martijnarts, line 270: Rewritten. Rules broken:

- 26-word sentence over the cap

````rust
/// Resolves every collected target against rust-analyzer's parse
///
/// The caller has already established that `file_id` belongs to a module
/// and that its recorded text matches the text the targets' byte offsets
/// were taken from.
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 14]
```diff
@@ -371,8 +519,50 @@ fn extract_result_error_type(display: &str) -> Option<String> {
...
+        let mut parser = tree_sitter::Parser::new();
+        parser
+            .set_language(&crate::language())
+            .expect("the Rust grammar should load");
+        let tree = parser.parse(source, None).expect("should parse");
+
+        DecoratedTree::new(tree, source.to_string(), file_path)
+    }
+
+    /// Exercises the exclusion guard, which no fixture workspace can reach
```
martijnarts, line 537: Rewritten. Rules broken:

- Test doc must be one or two sentences; this is a two-paragraph essay
- 30-word and 34-word sentences
- Rhetorical flourish ("its own test instead of trust", "cannot decline to have an opinion")

````rust
    /// An excluded file keeps its identity and loses everything else
    ///
    /// The VFS records such a file as present but holds neither its contents
    /// nor a source root for it, so every question the database can be asked
    /// about it panics. Whisker cannot decline to have an opinion: the
    /// exclusion is reported by the same [`Vfs::file_id`] call that hands
    /// back the id, so the guard is the only thing between that state and a
    /// panic raised from inside rust-analyzer.
    ///
    /// The state is seeded directly because the load whisker performs never
    /// produces it: rust-analyzer's language server marks files excluded
    /// from its own configuration, while Cargo's `[workspace] exclude` keeps
    /// a path out of the VFS entirely, without marking anything. That makes
    /// this the one guard no fixture workspace can exercise, which is
    /// exactly why it gets its own test instead of trust.
````

</details>

File: crates/whisker-rust/tests/fixtures/sample_project/orphan.rs [hunk 15]
```diff
@@ -0,0 +1,10 @@
+//! A file inside the package directory that no crate's module tree reaches
```
martijnarts: Rewritten. Rules broken:

- Em-dash aside pair ("— known to the toolchain, reachable by no crate —")
- -ing clause ("while resolving nothing about it")

````rust
//! A file inside the package directory that no crate's module tree reaches
//!
//! Cargo never compiles this file, because `src/lib.rs` declares no
//! `mod orphan`. rust-analyzer's VFS still interns it, because it sits
//! inside the package's source root. That combination — known to the
//! toolchain, reachable by no crate — is the case whisker used to report
//! clean while resolving nothing about it.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 16]
```diff
@@ -3,14 +3,49 @@ use std::sync::OnceLock;
...
-use whisker_types::{DecoratedNode, DecoratedTree, DecorationProvider};
+use whisker_types::{Coverage, CoverageGap, DecoratedNode, DecoratedTree, DecorationProvider};
 
 static PROVIDER: OnceLock<RustDecorationProvider> = OnceLock::new();
 
 fn fixture_path() -> PathBuf {
     Path::new(env!("CARGO_MANIFEST_DIR")).join("tests/fixtures/sample_project")
 }
 
+/// Writes a Cargo project whose crate reaches two files that are not UTF-8
```
martijnarts, line 14: Rewritten. Rules broken:

- Idiom ("on the fly", "take the whole test binary down")
- 41-word and 40-word sentences
- Passive with known agent ("is generated", "are reached")

````rust
/// Writes a Cargo project whose crate reaches two files that are not UTF-8
///
/// The project is generated on the fly because a committed file that is
/// not valid UTF-8 survives neither review nor most editors intact, and
/// because it must be a package of its own: rust-analyzer would otherwise
/// parse the bad module while building the definition map of whichever
/// crate owns it, and take the whole test binary down with it.
///
/// The two bad files are reached by the two different routes that make
/// rust-analyzer read a file's text. `src/bad.rs` is a module, read while
/// the definition map of the crate around it is built. `src/notes.md` is an
/// [`include_str!`] argument, read while the body that includes it is
/// inferred: a file whisker would never decorate, and whose extension is
/// therefore no reason to leave it broken.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 17]
```diff
@@ -66,6 +115,138 @@ fn find_first_node_of_kind<'a>(node: &DecoratedNode<'a>, kind: &str) -> Option<D
     None
 }
 
+/// Checks that a module that is not UTF-8 leaves its siblings covered
```
martijnarts, line 118: Rewritten. Rules broken:

- Test doc too long; the mechanism fits in two sentences
- 52-word sentence
- Idiom ("take the rest of its crate down with it", "turns this red")

````rust
/// One unreadable file must not take the rest of its crate down with it
///
/// `ra_ap_load_cargo` interns a file whose bytes are not UTF-8 without
/// recording its text, and rust-analyzer panics when it later reads that
/// text: while building the definition map of the crate around it, or
/// while inferring a body that includes it, so the file that dies is never
/// the one whisker was asked about. Before the load repaired such files,
/// this test aborted the whole binary with "Unable to fetch file text for
/// `vfs::FileId`" instead of failing. The fixture carries one of each
/// kind, so narrowing the repair back to Rust source turns this red too.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 18]
```diff
@@ -66,6 +115,138 @@ fn find_first_node_of_kind<'a>(node: &DecoratedNode<'a>, kind: &str) -> Option<D
...
+    let scrutinee = match_expr
+        .child_by_field_name("value")
+        .expect("match should have value field");
+    let ty = scrutinee
+        .decoration::<ResolvedType>()
+        .expect("scrutinee should have ResolvedType");
+    assert!(ty.is_enum(), "Color should be an enum");
+}
+
+/// Checks that code under `#[cfg(test)]` resolves like any other code
```
martijnarts, line 152: Rewritten. Rules broken:

- 28-word sentence
- "the difference between linting a project and linting the part that is not tested" is a rhetorical essay
- Significance inflation ("roughly half the lines of a Rust codebase")

````rust
/// Code inside `#[cfg(test)] mod tests` has to resolve like any other code
///
/// Cargo leaves `test` out of a crate's cfg options by default, which
/// resolves every such block to nothing while whisker still walks it and
/// reports the file clean. Roughly half the lines of a Rust codebase live
/// in those blocks, so this is the difference between linting a project and
/// linting the part of it that is not tested.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 19]
```diff
@@ -66,6 +115,138 @@ fn find_first_node_of_kind<'a>(node: &DecoratedNode<'a>, kind: &str) -> Option<D
...
+    let coverage = provider.decorate(&tree).expect("decorate should succeed");
+
+    match coverage {
+        Coverage::Covered(_) => panic!("a file outside the workspace must not be covered"),
+        Coverage::NotCovered(CoverageGap::OutsideWorkspace { .. }) => {}
+        Coverage::NotCovered(gap) => panic!("unexpected gap: {gap}"),
+    }
+}
+
+/// Checks that a generated file under the root reports
```
martijnarts, line 194: Rewritten. Rules broken:

- 33-word sentence
- Test doc too long; keep the pin and one mechanism
- Rhetorical flourish ("a remedy the user cannot act on")

````rust
/// A generated file under the root must not be called "outside" the root
///
/// `whisker check .` runs the workspace's build scripts, which create
/// `target`, so the next run offers the provider files under the workspace
/// root that rust-analyzer never interned. Reporting those as
/// [`CoverageGap::OutsideWorkspace`] prints a path plainly inside the root
/// next to a message claiming it is outside, and a remedy the user cannot
/// act on.
````

</details>

File: crates/whisker-types/src/decoration_provider.rs [hunk 20]
```diff
@@ -1,32 +1,54 @@
...
+/// decorate, and a provider that returned no decorations because it has
+/// never heard of the file, are indistinguishable by their output and must
+/// be distinguishable by their verdict; otherwise whisker reports a file
+/// clean that it could not read the meaning of.
 pub trait DecorationProvider: Send + Sync {
-    /// Populates decorations on the given tree
+    /// Returns the name this provider is identified by in diagnostics
+    fn name(&self) -> ProviderName;
+
+    /// Decorates the file, or explains why this provider cannot
```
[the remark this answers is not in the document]
martijnarts: You are right: the summary dropped the verb after "cannot". It now reads:

```rust
/// Decorates the file, or explains why the provider declines it
```

"Declines" is the word the rest of the `decorate` doc already uses ("a declined
file carries no partial state"). It also matches the summary on the
rust-analyzer implementation, which says the provider "declines the file".

I swept every doc summary this PR adds for the same defect and found two more
that were not complete clauses. The `RustDecorationProvider::NAME` constant
carried a passive noun phrase, "The name this provider is reported under in
diagnostics". It now reads "Identifies this provider in diagnostics". The
`orphan.rs` test fixture opened with a bare noun phrase; it now reads "Holds
code that no crate's module tree reaches". Every other summary this PR adds is
already a complete third-person clause with no trailing period.

The fixture doc also made a wrong claim, so I corrected it. It said Cargo skips
the file "because `src/lib.rs` declares no `mod orphan`". That is not the
reason. Rust resolves a `mod` declaration against the directory of the file that
holds it. So `mod orphan;` in `src/lib.rs` would look for `src/orphan.rs` and
fail with E0583.

Anyone who trusted the doc and added that declaration would break the fixture.
The doc now gives the real reason, which is that the file sits outside the
crate's `src` directory.

PR #216, opened 2026-08-05
Resolve types by span rather than by start offset

martijnarts:
The Rust provider asked rust-analyzer about a tree-sitter node by handing it
the node's first byte and taking the innermost expression around the token
found there. That answers a different question than the one being asked. The
first byte of `std::fs::read_to_string("x")` belongs to the token `std`, whose
innermost enclosing expression is the callee path, so the operand of `?` was
decorated with a function type, not with the `Result` the call produces. The
`anyhow_missing_context` rule bails on any operand that is not a `Result`,
which is why it had never once fired on the form it exists to catch.

The same fault reached match scrutinees, where `match palette.primary` was
typed as `palette` and `match pick_color()` as `pick_color`, so
`wildcard_match_arm` saw a struct and a function where an enum was sitting.
Every shape whose first token starts a smaller expression was affected: calls,
method calls, field access, indexing, and parentheses.

Each target now carries the full byte range of the syntax it is about, and the
lookup descends rust-analyzer's tree to the smallest node covering that range
before climbing to the enclosing expression. That reaches the expression the
range names whatever shape either parser gave it, without depending on the two
trees agreeing about what sits at one offset. Ranges that rust-analyzer's
32-bit offsets cannot express, and the zero-width nodes tree-sitter
synthesizes during error recovery, are dropped instead of resolved, because
both would name some unrelated expression.

The rules' own unit tests hand-build their decorations, so they went on
passing throughout. The provider's integration tests now run the real rules
over the real decorations the real provider makes, which is the only
arrangement that can tell a working rule from one whose input never arrives.
Those tests reach the rule crates through dev-dependencies that point back at
this crate; Cargo permits the cycle, and it costs nothing next to loading a
second workspace.

### DIFF
--- a/crates/whisker-rust/src/provider.rs
+++ b/crates/whisker-rust/src/provider.rs
@@ -7,8 +7,7 @@ use ra_ap_ide_db::base_db::SourceDatabase as _;
 use ra_ap_ide_db::{ChangeWithProcMacros, RootDatabase};
 use ra_ap_load_cargo::{LoadCargoConfig, ProcMacroServerChoice};
 use ra_ap_project_model::{CargoConfig, ProjectManifest, ProjectWorkspace};
-use ra_ap_syntax::AstNode;
-use ra_ap_syntax::ast;
+use ra_ap_syntax::{AstNode, NodeOrToken, SyntaxNode, TextRange, TextSize, ast};
 use ra_ap_vfs::{AbsPathBuf, FileExcluded, FileId, Vfs, VfsPath};
 use whisker_types::{
     Coverage, CoverageGap, DecoratedNode, DecoratedTree, DecorationMap, DecorationProvider,
@@ -207,11 +206,25 @@ impl DecorationProvider for RustDecorationProvider {
 
     /// Attaches type decorations to the tree's nodes, or declines the file
     ///
-    /// The provider covers a file only when four conditions hold. The VFS
-    /// must know the file. The toolchain must not exclude it. A crate
-    /// module must reach it. The tree's source must equal the text the
-    /// database holds. Each failed condition maps to a [`CoverageGap`]
-    /// variant.
+    /// Four conditions must hold before a single decoration is produced,
+    /// and their order matters. The file must be interned by the VFS, or
+    /// rust-analyzer has never heard of it. It must not be excluded, or
+    /// every question the database can be asked about it panics: the VFS
+    /// keeps an excluded file's identity and neither its contents nor a
+    /// source root, so the check has to come before the database is touched
+    /// at all, not merely before its text is read. It must belong to a
+    /// module, because `parse_guess_edition` papers over a missing module by
+    /// guessing the current edition and then resolves nothing. And the text
+    /// the pipeline parsed must equal the text the database holds, because
+    /// the byte ranges carried by every target index into rust-analyzer's
+    /// parse of that text.
+    ///
+    /// Only the first two conditions are guards. The module check protects
+    /// nothing, because establishing the module reads whatever sibling
+    /// modules the crate declares; a file whose text the load dropped would
+    /// already have brought the process down by then, which is why
+    /// [`give_text_to_files_that_have_none`] repairs the database at load
+    /// time, so the read here needs no careful ordering.
     ///
     /// # Errors
     ///
@@ -287,19 +300,16 @@ fn resolve_targets(
     let mut decorations = DecorationMap::new();
     for target in targets {
         match target {
-            Target::Function {
-                node_id,
-                start_byte,
-            } => {
-                if let Some(sig) = resolve_function(&ctx, *start_byte) {
+            Target::Function { node_id, range } => {
+                if let Some(sig) = resolve_function(&ctx, *range) {
                     decorations.insert(*node_id, sig);
                 }
             }
             Target::MatchScrutinee {
-                scrutinee_start_byte,
+                scrutinee_range,
                 scrutinee_node_id,
             } => {
-                if let Some((ty, flags)) = resolve_match_scrutinee(&ctx, *scrutinee_start_byte) {
+                if let Some((ty, flags)) = resolve_match_scrutinee(&ctx, *scrutinee_range) {
                     decorations.insert(*scrutinee_node_id, ty);
                     if let Some(flags) = flags {
                         decorations.insert(*scrutinee_node_id, flags);
@@ -307,18 +317,18 @@ fn resolve_targets(
                 }
             }
             Target::IfElse {
-                else_start_byte,
+                branch_range,
                 else_node_id,
             } => {
-                if let Some(ty) = resolve_expr_type(&ctx, *else_start_byte) {
+                if let Some(ty) = resolve_expr_type(&ctx, *branch_range) {
                     decorations.insert(*else_node_id, ty);
                 }
             }
             Target::TryExpr {
-                operand_start_byte,
+                operand_range,
                 operand_node_id,
             } => {
-                if let Some(ty) = resolve_expr_type(&ctx, *operand_start_byte) {
+                if let Some(ty) = resolve_expr_type(&ctx, *operand_range) {
                     decorations.insert(*operand_node_id, ty);
                 }
             }
@@ -327,37 +337,65 @@ fn resolve_targets(
     decorations
 }
 
+/// A node the provider asks rust-analyzer to resolve
+///
+/// Each variant pairs the byte range rust-analyzer inspects with the
+/// tree-sitter node ID that receives the decoration. These can differ: the
+/// provider types an `else` branch from its block but decorates the
+/// enclosing `else_clause`.
+///
+/// A full range, not a start offset, selects the node in rust-analyzer's
+/// tree. A start offset lands on one token, and the smallest expression
+/// around that token is often wrong. In `foo().bar()`, the start offset
+/// names the path `foo`, not the call.
 enum Target {
     Function {
         node_id: usize,
-        start_byte: usize,
+        range: TextRange,
     },
     MatchScrutinee {
-        scrutinee_start_byte: usize,
+        scrutinee_range: TextRange,
         scrutinee_node_id: usize,
     },
     IfElse {
-        else_start_byte: usize,
+        branch_range: TextRange,
         else_node_id: usize,
     },
     TryExpr {
-        operand_start_byte: usize,
+        operand_range: TextRange,
         operand_node_id: usize,
     },
+}
+
+/// Converts a tree-sitter byte range into a rust-analyzer [`TextRange`]
+///
+/// Rust-analyzer stores text offsets as 32 bits. When an offset does not
+/// fit, the function returns [`None`]. The caller then skips the target,
+/// because a truncated offset would name some other expression.
+fn text_range_of(node: &DecoratedNode<'_>) -> Option<TextRange> {
+    let range = node.raw().byte_range();
+    let start = u32::try_from(range.start).ok()?;
+    let end = u32::try_from(range.end).ok()?;
+
+    Some(TextRange::new(TextSize::from(start), TextSize::from(end)))
 }
 
 fn collect_targets(node: &DecoratedNode<'_>, targets: &mut Vec<Target>) {
     match node.kind() {
         "function_item" => {
-            targets.push(Target::Function {
-                node_id: node.id(),
-                start_byte: node.raw().start_byte(),
-            });
+            if let Some(range) = text_range_of(node) {
+                targets.push(Target::Function {
+                    node_id: node.id(),
+                    range,
+                });
+            }
         }
         "match_expression" => {
-            if let Some(scrutinee) = node.child_by_field_name("value") {
+            if let Some(scrutinee) = node.child_by_field_name("value")
+                && let Some(scrutinee_range) = text_range_of(&scrutinee)
+            {
                 targets.push(Target::MatchScrutinee {
-                    scrutinee_start_byte: scrutinee.raw().start_byte(),
+                    scrutinee_range,
                     scrutinee_node_id: scrutinee.id(),
                 });
             }
@@ -365,16 +403,20 @@ fn collect_targets(node: &DecoratedNode<'_>, targets: &mut Vec<Target>) {
         "if_expression" => {
             if let Some(alt) = node.child_by_field_name("alternative") {
                 let block = alt.named_child(0).unwrap_or(alt.clone());
-                targets.push(Target::IfElse {
-                    else_start_byte: block.raw().start_byte(),
-                    else_node_id: alt.id(),
-                });
+                if let Some(branch_range) = text_range_of(&block) {
+                    targets.push(Target::IfElse {
+                        branch_range,
+                        else_node_id: alt.id(),
+                    });
+                }
             }
         }
         "try_expression" => {
-            if let Some(operand) = node.named_child(0) {
+            if let Some(operand) = node.named_child(0)
+                && let Some(operand_range) = text_range_of(&operand)
+            {
                 targets.push(Target::TryExpr {
-                    operand_start_byte: operand.raw().start_byte(),
+                    operand_range,
                     operand_node_id: operand.id(),
                 });
             }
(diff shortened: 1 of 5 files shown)


martijnarts:
Every doc comment this PR adds was re-checked against three rule sets and rewritten where it failed. 11 blocks changed here.

Violations found here: 7 reviewer verdicts from #208, 19 ASD-STE100, 6 AI-writing tells.

|---|---|---|
| `crates/whisker-rust/src/provider.rs` | enum Target | 29-word sentence (else_clause sentence); 44-word sentence (foo().bar() sentence); Passive 'is asked' with known agent; 'The range is a range' redundancy, wordy for the item |
| `crates/whisker-rust/src/provider.rs` | fn covering_node | 45-word sentence ('Descending to the smallest covering node...'); 32-word sentence (zero-width sentence); Passive 'is refused' with known agent; Omits that an out-of-tree range also retur... |
| `crates/whisker-rust/src/provider.rs` | fn text_range_of | Never states that overflow returns None (the WHAT); 27-word sentence; -ing subject 'Dropping the target'; Figurative 'the only honest answer', 'quietly' |
| `crates/whisker-rust/tests/provider_integration.rs` | fn flagged_within | 31-word sentence; Passive 'are attributed' with known agent; -ing subject 'Returning the covered text' |
| `crates/whisker-rust/tests/provider_integration.rs` | test anyhow_missing_context_flags_a_bare_try_in_an_anyhow_function | Test doc too long, four sentences; 27-word sentence; -ing subject 'Running the real provider'; 'the only arrangement that can tell the two apart' inflation |
| `crates/whisker-rust/tests/provider_integration.rs` | test match_scrutinee_on_call_expression_has_is_enum_true | -ing subject 'Typing that' where a finite verb works |
| `crates/whisker-rust/tests/provider_integration.rs` | test match_scrutinee_on_field_expression_has_is_enum_true | Test doc too long, not parseable at a glance; 33-word sentence; 'quietly' significance inflation |
| `crates/whisker-rust/tests/provider_integration.rs` | test try_operand_on_call_expression_is_typed_as_the_call_result | Test doc too long; 35-word sentence; 'gives up before it looks at anything else' figure of speech |
| `crates/whisker-rust/tests/provider_integration.rs` | test try_operand_on_method_call_is_typed_as_the_call_result | Passive 'it is pinned' with known agent |
| `crates/whisker-rust/tests/provider_integration.rs` | test wildcard_match_arm_flags_a_scrutinee_reached_through_a_field | 28-word sentence; Idiom 'got right'; Summary says the rule 'reaches' scrutinees; the provider does |
| `crates/whisker-rust/tests/provider_integration.rs` | test wildcard_match_arm_ignores_a_scrutinee_that_is_not_an_enum | Idiom 'the cheapest place to notice that'; -ing subject in summary ('Reaching more scrutinees') |

crates/whisker-rust/src/provider.rs
```diff
@@ -327,54 +337,86 @@ fn resolve_targets(
     decorations
 }
 
+/// A node the provider asks rust-analyzer to resolve
```
martijnarts, line 340: Rewritten. Rules broken:

- 29-word sentence (else_clause sentence)
- 44-word sentence (foo().bar() sentence)
- Passive 'is asked' with known agent
- 'The range is a range' redundancy, wordy for the item

````rust
/// A node whose type or signature the provider will ask rust-analyzer about
///
/// Each variant carries the byte range of the syntax the question is about
/// and the tree-sitter node ID the answer belongs on. The two are not always
/// the same node: an `else` branch is typed from its block but decorated on
/// the enclosing `else_clause`, which is the node lints reach for.
///
/// The range is a range and not a start offset because rust-analyzer's tree
/// is asked which of its nodes the range covers. An offset only identifies a
/// token, and the innermost expression around a token is routinely not the
/// expression the range names: the offset of `foo().bar()` lands on `foo`,
/// whose innermost enclosing expression is the path `foo`, typed as a
/// function, not as the call's result.
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 2]
```diff
@@ -327,54 +337,86 @@ fn resolve_targets(
...
         else_node_id: usize,
     },
     TryExpr {
-        operand_start_byte: usize,
+        operand_range: TextRange,
         operand_node_id: usize,
     },
 }
 
+/// Converts a tree-sitter byte range into a rust-analyzer [`TextRange`]
```
martijnarts, line 370: Rewritten. Rules broken:

- Never states that overflow returns None (the WHAT)
- 27-word sentence
- -ing subject 'Dropping the target'
- Figurative 'the only honest answer', 'quietly'

````rust
/// Converts a tree-sitter node's byte range into rust-analyzer's
///
/// Rust-analyzer addresses text with 32-bit offsets, so a file too large for
/// one is a file it cannot describe. Dropping the target is then the only
/// honest answer, because a truncated offset would quietly name some other
/// expression and decorate this node with that expression's type.
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 3]
```diff
@@ -400,29 +442,42 @@ impl DecorateCtx<'_, '_> {
...
-fn find_enclosing_fn(
-    ctx: &DecorateCtx<'_, '_>,
-    offset: ra_ap_syntax::TextSize,
-) -> Option<ra_ap_hir::Function> {
-    let fn_node = ctx
-        .syntax
-        .token_at_offset(offset)
-        .right_biased()?
-        .parent_ancestors()
+/// Returns the smallest node in rust-analyzer's tree that covers `range`
```
martijnarts, line 445: Rewritten. Rules broken:

- 45-word sentence ('Descending to the smallest covering node...')
- 32-word sentence (zero-width sentence)
- Passive 'is refused' with known agent
- Omits that an out-of-tree range also returns None

````rust
/// Returns the smallest node in rust-analyzer's tree that covers `range`
///
/// The two parsers agree on byte offsets, because the caller has already
/// checked that they were handed the same text, but not on tree shape. A
/// tree-sitter node is therefore located by the span it occupies, not by
/// position in the tree. Descending to the smallest covering node and then
/// climbing is what makes the lookup shape-independent: whichever node
/// rust-analyzer happens to have at that span, the ancestor walk from it
/// reaches the smallest [`ast::Expr`] or [`ast::Fn`] enclosing the whole
/// span, which is the one the caller meant.
///
/// An empty range is refused. Tree-sitter synthesizes zero-width nodes while
/// recovering from a parse error, and a zero-width span "covers" whatever
/// token it sits inside, which would decorate the broken node with an
/// unrelated neighbor's type.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 4]
```diff
@@ -115,6 +119,32 @@ fn find_first_node_of_kind<'a>(node: &DecoratedNode<'a>, kind: &str) -> Option<D
     None
 }
 
+/// Returns the operand of the first `?` expression inside `node`
+fn find_first_try_operand<'a>(node: &DecoratedNode<'a>) -> Option<DecoratedNode<'a>> {
+    find_first_node_of_kind(node, "try_expression")?.named_child(0)
+}
+
+/// Returns the source text covered by each diagnostic raised inside `func`
```
martijnarts, line 129: Rewritten. Rules broken:

- 31-word sentence
- Passive 'are attributed' with known agent
- -ing subject 'Returning the covered text'

````rust
/// Returns the source text covered by each diagnostic raised inside `func`
///
/// Diagnostics are attributed to a function by byte containment, not by
/// their position in the list, because these tests are about which of the
/// fixture's functions a lint reaches at all. An expectation written against
/// list positions would survive a lint that stopped seeing one function and
/// started seeing another.
///
/// Returning the covered text instead of raw offsets also lets the fixture
/// grow without every expectation being renumbered.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 5]
```diff
@@ -362,6 +392,70 @@ fn match_scrutinee_on_integer_has_is_enum_false() {
     assert!(!ty.is_enum(), "i32 should not be an enum");
 }
 
+/// A scrutinee reached through a field must be typed as the field
```
martijnarts, line 397: Rewritten. Rules broken:

- Test doc too long, not parseable at a glance
- 33-word sentence
- 'quietly' significance inflation

````rust
/// A scrutinee reached through a field must be typed as the field
///
/// `match palette.primary` starts on the token `palette`, and the smallest
/// expression around that token is the path `palette`, whose type is the
/// struct. Typing the scrutinee from where it starts, instead of from what
/// it spans, therefore reports a struct, and every lint that asks whether the
/// scrutinee is an enum quietly gets the wrong answer.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 6]
```diff
@@ -362,6 +392,70 @@ fn match_scrutinee_on_integer_has_is_enum_false() {
...
+
+    assert_eq!(scrutinee.text(), "palette.primary");
+    assert!(
+        ty.is_enum(),
+        "the field's Color type should be an enum, got: {}",
+        ty.display()
+    );
+}
+
+/// A scrutinee returned by a call must be typed as the call's result
```
martijnarts, line 429: Rewritten. Rules broken:

- -ing subject 'Typing that' where a finite verb works

````rust
/// A scrutinee returned by a call must be typed as the call's result
///
/// `match pick_color()` starts on the token `pick_color`, whose smallest
/// enclosing expression is the callee path. Typing that yields the function
/// itself, not the `Color` it returns.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 7]
```diff
@@ -437,6 +531,158 @@ fn if_let_with_non_diverging_else_is_not_never() {
     }
 }
 
+/// The operand of `?` must be typed as the call, not as the callee
```
martijnarts, line 536: Rewritten. Rules broken:

- Test doc too long
- 35-word sentence
- 'gives up before it looks at anything else' figure of speech

````rust
/// The operand of `?` must be typed as the call, not as the callee
///
/// `std::fs::read_to_string("anyhow_bare.txt")?` starts on the token `std`,
/// and the smallest expression around that token is the callee path. Typing
/// the operand from where it starts therefore reports a function type, which
/// is not a `Result`, and every lint gated on the operand being a `Result`
/// gives up before it looks at anything else.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 8]
```diff
@@ -437,6 +531,158 @@ fn if_let_with_non_diverging_else_is_not_never() {
...
+        "std::fs::read_to_string(\"anyhow_bare.txt\")"
+    );
+    assert!(
+        ty.is_result(),
+        "read_to_string returns a Result, got: {}",
+        ty.display()
+    );
+}
+
+/// The operand of `?` must be typed as the method call, not as the receiver
```
martijnarts, line 567: Rewritten. Rules broken:

- Passive 'it is pinned' with known agent

````rust
/// The operand of `?` must be typed as the method call, not as the receiver
///
/// `loader.load()?` starts on the token `loader`, whose type is a reference
/// to the struct. This is the same fault as the callee case, reached through
/// a different shape, so it is pinned separately.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 9]
```diff
@@ -437,6 +531,158 @@ fn if_let_with_non_diverging_else_is_not_never() {
...
+
+    assert_eq!(operand.text(), "loader.load()");
+    assert!(
+        ty.is_result(),
+        "Loader::load returns a Result, got: {}",
+        ty.display()
+    );
+}
+
+/// The `anyhow_missing_context` rule fires on decorations the provider made
```
martijnarts, line 617: Rewritten. Rules broken:

- Test doc too long, four sentences
- 27-word sentence
- -ing subject 'Running the real provider'
- 'the only arrangement that can tell the two apart' inflation

````rust
/// The `anyhow_missing_context` rule fires on decorations the provider made
///
/// Its own unit tests attach the decorations by hand, so they went on passing
/// for as long as the provider produced decorations that never satisfied the
/// rule. Running the real provider and the real rule over real source is the
/// only arrangement that can tell the two apart.
///
/// The expectation is the complete list of what the rule flags inside the
/// function, so it also pins the negative: the `?` on the neighboring
/// `.context(..)` call must not be reported.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 10]
```diff
@@ -437,6 +531,158 @@ fn if_let_with_non_diverging_else_is_not_never() {
...
+
+    let diagnostics = whisker_core::walk(&tree, &mut passes);
+
+    assert_eq!(
+        flagged_within(&tree, &diagnostics, &func),
+        vec!["loader.load()?"]
+    );
+}
+
+/// The `wildcard_match_arm` rule fires on a scrutinee reached through a field
```
martijnarts, line 657: Rewritten. Rules broken:

- 28-word sentence
- Idiom 'got right'
- Summary says the rule 'reaches' scrutinees; the provider does

````rust
/// The `wildcard_match_arm` rule sees scrutinees it reaches through a field
///
/// The rule was already firing on the fixture, but only on scrutinees that
/// are a bare name, which is the one shape the old start-offset lookup got
/// right. This pins the shapes it used to miss.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 11]
```diff
@@ -437,6 +531,158 @@ fn if_let_with_non_diverging_else_is_not_never() {
...
+        .expect("should find match_on_call_expression");
+    let mut passes: Vec<Box<dyn LintPass>> =
+        vec![Box::new(RustLintPassAdapter::new(WildcardMatchArm))];
+
+    let diagnostics = whisker_core::walk(&tree, &mut passes);
+
+    assert_eq!(flagged_within(&tree, &diagnostics, &func), vec!["_"]);
+}
+
+/// The `wildcard_match_arm` rule still ignores a non-enum scrutinee
```
martijnarts, line 691: Rewritten. Rules broken:

- Idiom 'the cheapest place to notice that'
- -ing subject in summary ('Reaching more scrutinees')

````rust
/// Reaching more scrutinees must not mean reaching every scrutinee
///
/// A lookup that resolved to something too broad would make `is_enum` true
/// far too often, and this is the cheapest place to notice that.
````

</details>

**Review by martijnarts — COMMENTED** [review]

File: crates/whisker-rust/tests/provider_integration.rs [hunk 12]
```diff
@@ -118,6 +122,36 @@ fn find_first_node_of_kind<'a>(node: &DecoratedNode<'a>, kind: &str) -> Option<D
...
+    tree: &'a DecoratedTree,
+    diagnostics: &[Diagnostic],
+    func: &DecoratedNode<'_>,
+) -> Vec<&'a str> {
+    let range = func.raw().byte_range();
+
+    diagnostics
+        .iter()
+        .map(Diagnostic::span)
+        .filter(|span| range.contains(&span.start()))
```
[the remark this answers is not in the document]
martijnarts: You are right. `flagged_within` now requires full containment:

```rust
let Range { start, end } = func.raw().byte_range();

diagnostics
    .iter()
    .map(Diagnostic::span)
    .filter(|span| start <= span.start() && span.end() <= end)
```

I tightened the helper rather than relax its doc. Each rule-driven caller
expects the complete list of what one rule flags inside one function. With
start containment, a span that runs past the closing brace still lands in an
expectation that claims to be exhaustive.

No existing test changes its outcome. Every diagnostic these tests see comes
from `DecoratedNode::span`, and tree-sitter nodes nest, so a node that starts
inside a function also ends inside it. I ran the integration tests both ways.
With full containment all 27 pass. With start containment the 26 older tests
pass and only the new test fails.

That new test is
`flagged_within_with_a_span_past_the_function_end_returns_nothing`. It builds a
span from the start of `match_on_enum` to one byte past its end, and asserts
the helper drops it. No lint produces that shape, so the test builds the span
by hand.

crates/whisker-rust/src/provider.rs
```diff
@@ -377,54 +373,89 @@ fn resolve_targets(
...
         else_node_id: usize,
     },
     TryExpr {
-        operand_start_byte: usize,
+        operand_range: TextRange,
         operand_node_id: usize,
     },
 }
 
+/// Converts a tree-sitter node's byte range into rust-analyzer's
```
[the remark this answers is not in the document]
martijnarts: You were right. This comment sits against `da50f1f`, where the summary broke
off at "into rust-analyzer's". I finished the sentence in a later
documentation pass, so `text_range_of` now reads:

```rust
/// Converts a tree-sitter byte range into a rust-analyzer [`TextRange`]
///
/// Rust-analyzer stores text offsets as 32 bits. When an offset does not
/// fit, the function returns [`None`]. The caller then skips the target,
/// because a truncated offset would name some other expression.
```

## Sweep — crates/whisker-rust/Cargo.toml
VERDICT: fixed
The comment on the dev-dependency cycle carried a 29-word sentence. It now
reads:

```toml
# The lint crates depend on this one, so the two lint entries below close a
# cycle. Cargo permits that for dev-dependencies. The cycle lets the
# integration tests assert on diagnostics that real lints draw from real
# decorations, not on the decorations alone.
```

"These edges" became "the two lint entries below", because `proptest` and
`whisker-core` sit in the same block and neither closes a cycle.

## Sweep — doc on wildcard_match_arm_flags_a_scrutinee_reached_through_a_field
VERDICT: deferred
The doc on that test reads:

```rust
/// The old start-offset lookup only resolved scrutinees that are a bare
/// name. This test pins a shape that lookup used to miss.
```

That narrates the change rather than the test. It is also broader than the
truth, because a start offset resolves a literal scrutinee correctly too.

I left it alone because you wrote that replacement yourself in the last doc
pass, so it is your call. Here is a version without the history:

```rust
/// A field scrutinee resolves only when the lookup spans the whole
/// expression.
```

PR #217, opened 2026-08-05
Add types that name a Rust item rather than render it

martijnarts:
`anyhow_missing_context` decides whether a function returns an anyhow error by
comparing the rendered type name against `"anyhow::Error"` or `"Error"`.
rust-analyzer renders `syn::Error`, `serde_json::Error`, `std::io::Error`, and
`std::fmt::Error` all as `Error`, so the comparison cannot tell them apart. A
display string does not carry type identity, and no amount of string
comparison recovers it.

This adds the types that carry it instead. `TypePath` names the item that
defines a type (crate, module path, item), and `TypePathRef` is its borrowed
form, so a lint can hold a path as a constant and compare against it without
allocating. `ErrorType` pairs a function's error type with that path.
`ReturnMode` records whether a signature was read directly, awaited through an
`async fn`, or left opaque, which keeps "async and unprojectable"
distinguishable from "async and not a `Result`".

None of it is wired up here. The change that teaches the provider to resolve
identities and the rule to compare them comes right after.

### DIFF
--- a/crates/whisker-rust/src/decorations/error_type.rs
+++ b/crates/whisker-rust/src/decorations/error_type.rs
@@ -1,0 +1,111 @@
+use crate::decorations::{TypePath, TypePathRef};
+
+/// The `E` of the [`Result<T, E>`] a function returns
+///
+/// Not every `E` has a defining item. A type parameter, a reference, or a
+/// tuple has none, so no [`TypePath`] exists for it.
+///
+/// `Box<dyn Error>` is [`ErrorType::Named`] with the path
+/// `alloc::boxed::Box`, because `Box` is the ADT in the `E` slot.
+///
+/// [`Result<T, E>`]: std::result::Result
+/// [`TypePath`]: crate::decorations::TypePath
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub enum ErrorType {
+    /// The error is an ADT, identified by where it is defined
+    Named(TypePath),
+    /// The error is a type parameter, not a concrete type
+    Generic,
+    /// The error is a type with no defining item
+    Unnamed,
+}
+
+impl ErrorType {
+    /// Returns whether the error type is the one `path` names
+    ///
+    /// The comparison uses definition paths, not rendered names, so
+    /// `syn::Error` does not match `anyhow::Error`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use whisker_rust::decorations::{ErrorType, TypePath, TypePathRef};
+    ///
+    /// const ANYHOW_ERROR: TypePathRef<'static> = TypePathRef::new("anyhow", &[], "Error");
+    ///
+    /// let error = ErrorType::Named(TypePath::new("syn", ["error"], "Error"));
+    ///
+    /// assert!(!error.is(ANYHOW_ERROR));
+    /// ```
+    pub fn is(&self, path: TypePathRef<'_>) -> bool {
+        match self {
+            ErrorType::Named(actual) => *actual == path,
+            ErrorType::Generic => false,
+            ErrorType::Unnamed => false,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    const ANYHOW_ERROR: TypePathRef<'static> = TypePathRef::new("anyhow", &[], "Error");
+
+    #[test]
+    fn is_with_generic_returns_false() {
+        let error = ErrorType::Generic;
+
+        let matches = error.is(ANYHOW_ERROR);
+
+        assert!(!matches);
+    }
+
+    #[test]
+    fn is_with_matching_path_returns_true() {
+        let error = ErrorType::Named(TypePath::new("anyhow", [] as [&str; 0], "Error"));
+
+        let matches = error.is(ANYHOW_ERROR);
+
+        assert!(matches);
+    }
+
+    /// Pins that `syn::Error` does not match a path that names `anyhow::Error`
+    ///
+    /// Both render as `Error`, so only the definition path distinguishes them.
+    #[test]
+    fn is_with_same_name_in_another_crate_returns_false() {
+        let error = ErrorType::Named(TypePath::new("syn", ["error"], "Error"));
+
+        let matches = error.is(ANYHOW_ERROR);
+
+        assert!(!matches);
+    }
+
+    #[test]
+    fn is_with_unnamed_returns_false() {
+        let error = ErrorType::Unnamed;
+
+        let matches = error.is(ANYHOW_ERROR);
+
+        assert!(!matches);
+    }
+
+    #[test]
+    fn trait_send_error_type() {
+        fn assert_send<T: Send>() {}
+        assert_send::<ErrorType>();
+    }
+
+    #[test]
+    fn trait_sync_error_type() {
+        fn assert_sync<T: Sync>() {}
+        assert_sync::<ErrorType>();
+    }
+
+    #[test]
+    fn trait_unpin_error_type() {
+        fn assert_unpin<T: Unpin>() {}
+        assert_unpin::<ErrorType>();
+    }
+}
--- a/crates/whisker-rust/src/decorations/return_mode.rs
+++ b/crates/whisker-rust/src/decorations/return_mode.rs
@@ -1,0 +1,37 @@
+/// Where a function's error type comes from
+///
+/// For a plain function, the declared return type holds the error type. For
+/// an `async fn`, the declared type is an opaque future, and the error type
+/// is in the future's output.
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub enum ReturnMode {
+    /// The error type is in the declared return type
+    Direct,
+    /// The function is `async`; the error type is in the future's output
+    Awaited,
+    /// The function is `async`, but the future's output is unknown
+    Opaque,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn trait_send_return_mode() {
+        fn assert_send<T: Send>() {}
+        assert_send::<ReturnMode>();
+    }
+
+    #[test]
+    fn trait_sync_return_mode() {
+        fn assert_sync<T: Sync>() {}
+        assert_sync::<ReturnMode>();
+    }
+
+    #[test]
+    fn trait_unpin_return_mode() {
+        fn assert_unpin<T: Unpin>() {}
+        assert_unpin::<ReturnMode>();
+    }
+}
--- a/crates/whisker-rust/src/decorations/type_path.rs
+++ b/crates/whisker-rust/src/decorations/type_path.rs
@@ -1,0 +1,210 @@
+use std::fmt;
+
+use crate::decorations::TypePathRef;
+
+/// Where a type is defined, as a path from its crate root
+///
+/// Rust-analyzer can render `syn::Error`, `std::io::Error`, and
+/// `anyhow::Error` all as `Error`. A rule that must distinguish them needs
+/// the definition, not the rendered name.
+///
+/// The path names the item's definition, not a re-export, so `syn::Error`
+/// appears here as `syn::error::Error`.
+///
+/// # Examples
+///
+/// ```
+/// use whisker_rust::decorations::TypePath;
+///
+/// let path = TypePath::new("syn", ["error"], "Error");
+///
+/// assert_eq!(path.to_string(), "syn::error::Error");
+/// ```
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub struct TypePath {
+    krate: Box<str>,
+    modules: Box<[Box<str>]>,
+    name: Box<str>,
+}
+
+impl TypePath {
+    /// Creates a path from a crate name, its module segments, and an item name
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use whisker_rust::decorations::TypePath;
+    ///
+    /// let path = TypePath::new("anyhow", [] as [&str; 0], "Error");
+    ///
+    /// assert_eq!(path.to_string(), "anyhow::Error");
+    /// ```
+    pub fn new<M>(krate: &str, modules: M, name: &str) -> Self
+    where
+        M: IntoIterator,
+        M::Item: AsRef<str>,
+    {
+        Self {
+            krate: Box::from(krate),
+            modules: modules
+                .into_iter()
+                .map(|segment| Box::from(segment.as_ref()))
+                .collect(),
+            name: Box::from(name),
+        }
+    }
+
+    /// Returns the name of the crate that defines the type
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use whisker_rust::decorations::TypePath;
+    ///
+    /// let path = TypePath::new("syn", ["error"], "Error");
+    ///
+    /// assert_eq!(path.krate(), "syn");
+    /// ```
+    pub fn krate(&self) -> &str {
+        &self.krate
+    }
+
+    /// Returns the module segments between the crate root and the definition
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use whisker_rust::decorations::TypePath;
+    ///
+    /// let path = TypePath::new("syn", ["error"], "Error");
+    ///
+    /// assert_eq!(path.modules().collect::<Vec<_>>(), vec!["error"]);
+    /// ```
+    pub fn modules(&self) -> impl Iterator<Item = &str> {
+        self.modules.iter().map(AsRef::as_ref)
+    }
+
+    /// Returns the item's own name
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use whisker_rust::decorations::TypePath;
+    ///
+    /// let path = TypePath::new("syn", ["error"], "Error");
+    ///
+    /// assert_eq!(path.name(), "Error");
+    /// ```
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+}
+
+/// Renders the crate, the module segments, and the name, joined by `::`
+///
+/// To test identity, compare the path against a [`TypePathRef`].
+///
+/// [`TypePathRef`]: crate::decorations::TypePathRef
+impl fmt::Display for TypePath {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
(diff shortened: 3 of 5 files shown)


martijnarts:
Every doc comment this PR adds was re-checked against three rule sets and rewritten where it failed. 13 blocks changed here.

Violations found here: 1 stated something untrue at this branch, 7 reviewer verdicts from #208, 18 ASD-STE100, 4 AI-writing tells.

|---|---|---|
| `crates/whisker-rust/src/decorations/error_type.rs` | ErrorType | Wordy explanation for a simple enum; meta commentary ('Saying so is better than inventing a name', 'That is factually right ... will surprise a reader'); 41-word sentence ('A type paramet... |
| `crates/whisker-rust/src/decorations/error_type.rs` | ErrorType::Generic | Idiom 'pin down' |
| `crates/whisker-rust/src/decorations/error_type.rs` | ErrorType::Unnamed | Vague word 'something'; dangling 'to name' |
| `crates/whisker-rust/src/decorations/error_type.rs` | ErrorType::is | Figure of speech 'never answers to'; Passive 'The comparison is between definitions rust-analyzer resolved' |
| `crates/whisker-rust/src/decorations/error_type.rs` | tests::is_with_same_name_in_another_crate_returns_false | Test docs must be short and parseable; Significance inflation 'This is the whole point of the type'; Idiom 'answer to' |
| `crates/whisker-rust/src/decorations/return_mode.rs` | ReturnMode | Paragraph about what a test can assert is unrelated to the item; simple enum, wordy doc; 31-word sentence ('Recording which of the two was read means ...'); '-ing' opener 'Recording'; pas... |
| `crates/whisker-rust/src/decorations/return_mode.rs` | ReturnMode::Awaited | Figure of speech '`?` sees' |
| `crates/whisker-rust/src/decorations/return_mode.rs` | ReturnMode::Direct | Figure of speech '`?` sees' |
| `crates/whisker-rust/src/decorations/return_mode.rs` | ReturnMode::Opaque | Passive 'could not be projected', 'nothing is known'; Figure of speech 'what `?` sees' |
| `crates/whisker-rust/src/decorations/type_path.rs` | TypePath | Essay on rejected design alternatives (trait check, .context() check) is unrelated to the item; Filler 'it is worth saying why'; near-aside 'Nominal identity is the honest primitive'; 40-... |
| `crates/whisker-rust/src/decorations/type_path.rs` | impl Display for TypePath | Figure of speech 'cannot be fooled'; passive |
| `crates/whisker-rust/src/decorations/type_path.rs` | tests::eq_with_different_module_depth_is_false | Idioms 'pass for', 'answer to'; '-ing' opener 'Zipping'; Test docs must be short and parseable |
| `crates/whisker-rust/src/decorations/type_path_ref.rs` | TypePathRef | Verbless fragment 'Const-constructible so ...'; 33-word sentence; Rather-than (twice: 'instead of open-coding', 'rather than `TypeRef`'); Naming-rationale essay unrelated to what the item... |

- `crates/whisker-rust/src/decorations/return_mode.rs` &mdash; ReturnMode: No provider records ReturnMode at this branch, so the test-assertion story documents nothing that exists

crates/whisker-rust/src/decorations/error_type.rs
```diff
@@ -0,0 +1,111 @@
+use crate::decorations::{TypePath, TypePathRef};
+
+/// The `E` of the [`Result<T, E>`] a function returns
```
martijnarts, line 3: Rewritten. Rules broken:

- Wordy explanation for a simple enum; meta commentary ('Saying so is better than inventing a name', 'That is factually right ... will surprise a reader')
- 41-word sentence ('A type parameter, a reference, and a tuple are all types ...')
- Passive 'could be pointed at'
- Significance inflation in the Box<dyn Error> paragraph

````rust
/// The `E` of the [`Result<T, E>`] a function returns
///
/// Not every `E` has a name. A type parameter, a reference, and a tuple are
/// all types rust-analyzer resolved perfectly well, and none of them is
/// defined by an item that could be pointed at, so there is nothing to
/// compare a [`TypePath`] against. Saying so is better than inventing a name
/// for them.
///
/// `Box<dyn Error>` is [`ErrorType::Named`], not something dynamic: the `E`
/// slot holds `Box`, which is an ADT, so its identity is `alloc::boxed::Box`.
/// That is factually right and produces the right lint outcome, but it will
/// surprise a reader expecting a trait-object variant. An unsized
/// `dyn Error` cannot occupy the `E` slot at all, which is why there is no
/// such variant.
///
/// [`Result<T, E>`]: std::result::Result
/// [`TypePath`]: crate::decorations::TypePath
````

</details>

File: crates/whisker-rust/src/decorations/error_type.rs [hunk 2]
```diff
@@ -0,0 +1,111 @@
...
+/// `Box<dyn Error>` is [`ErrorType::Named`] with identity `alloc::boxed::Box`,
+/// because `Box` is the ADT in the `E` slot.
+///
+/// [`Result<T, E>`]: std::result::Result
+/// [`TypePath`]: crate::decorations::TypePath
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub enum ErrorType {
+    /// The error is an ADT, identified by where it is defined
+    Named(TypePath),
+    /// The error is a type parameter, not a concrete type
```
martijnarts, line 17: Rewritten. Rules broken:

- Idiom 'pin down'

````rust
    /// The error is a type parameter the signature does not pin down
````

</details>

File: crates/whisker-rust/src/decorations/error_type.rs [hunk 3]
```diff
@@ -0,0 +1,111 @@
...
+    Generic,
+    /// The error is a type with no defining item
```
martijnarts, line 19: Rewritten. Rules broken:

- Vague word 'something'; dangling 'to name'

````rust
    /// The error resolved to something with no defining item to name
````

</details>

File: crates/whisker-rust/src/decorations/error_type.rs [hunk 4]
```diff
@@ -0,0 +1,111 @@
...
+    Unnamed,
+}
+
+impl ErrorType {
+    /// Returns whether the error type is the one `path` names
```
martijnarts, line 24: Rewritten. Rules broken:

- Figure of speech 'never answers to'
- Passive 'The comparison is between definitions rust-analyzer resolved'

````rust
    /// Returns whether the error type is the one `path` names
    ///
    /// The comparison is between definitions rust-analyzer resolved, not
    /// names it printed, so a crate's own `Error` never answers to
    /// `anyhow::Error`.
    ///
    /// # Examples
    ///
    /// ```
    /// use whisker_rust::decorations::{ErrorType, TypePath, TypePathRef};
    ///
    /// const ANYHOW_ERROR: TypePathRef<'static> = TypePathRef::new("anyhow", &[], "Error");
    ///
    /// let error = ErrorType::Named(TypePath::new("syn", ["error"], "Error"));
    ///
    /// assert!(!error.is(ANYHOW_ERROR));
    /// ```
````

</details>

File: crates/whisker-rust/src/decorations/error_type.rs [hunk 5]
```diff
@@ -0,0 +1,111 @@
...
+    #[test]
+    fn is_with_matching_path_returns_true() {
+        let error = ErrorType::Named(TypePath::new("anyhow", [] as [&str; 0], "Error"));
+
+        let matches = error.is(ANYHOW_ERROR);
+
+        assert!(matches);
+    }
+
+    /// Pins that `syn::Error` does not match a path that names `anyhow::Error`
```
martijnarts, line 73: Rewritten. Rules broken:

- Test docs must be short and parseable
- Significance inflation 'This is the whole point of the type'
- Idiom 'answer to'

````rust
    /// A crate's own `Error` must not answer to anyhow's
    ///
    /// This is the whole point of the type: `syn::Error`, `std::io::Error`,
    /// and `anyhow::Error` all render as the bare word `Error`, so a rule
    /// comparing renderings flags all three.
````

</details>

File: crates/whisker-rust/src/decorations/return_mode.rs [hunk 6]
```diff
@@ -0,0 +1,37 @@
+/// Where a function's error type comes from
```
martijnarts, line 1: Rewritten. Rules broken:

- Paragraph about what a test can assert is unrelated to the item; simple enum, wordy doc
- 31-word sentence ('Recording which of the two was read means ...')
- '-ing' opener 'Recording'; passive 'had to be read', 'was read'
- No provider records ReturnMode at this branch, so the test-assertion story documents nothing that exists

````rust
/// Where a function's error type had to be read from
///
/// An `async fn`'s written return type is the opaque future, not the type
/// `?` in its body converts into. Recording which of the two was read means
/// a test can assert that the provider went through the future instead of
/// asserting a result a provider ignoring `async` could also produce.
````

</details>

File: crates/whisker-rust/src/decorations/return_mode.rs [hunk 7]
```diff
@@ -0,0 +1,37 @@
...
+///
+/// For a plain function, the declared return type holds the error type. For
+/// an `async fn`, the declared type is an opaque future, and the error type
+/// is in the future's output.
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub enum ReturnMode {
+    /// The error type is in the declared return type
```
martijnarts, line 8: Rewritten. Rules broken:

- Figure of speech '`?` sees'

````rust
    /// `?` sees the declared return type
````

</details>

File: crates/whisker-rust/src/decorations/return_mode.rs [hunk 8]
```diff
@@ -0,0 +1,37 @@
...
+    Direct,
+    /// The function is `async`; the error type is in the future's output
```
martijnarts, line 10: Rewritten. Rules broken:

- Figure of speech '`?` sees'

````rust
    /// The function is `async`; `?` sees the future's output
````

</details>

File: crates/whisker-rust/src/decorations/return_mode.rs [hunk 9]
```diff
@@ -0,0 +1,37 @@
...
+    Awaited,
+    /// The function is `async`, but the future's output is unknown
```
martijnarts, line 12: Rewritten. Rules broken:

- Passive 'could not be projected', 'nothing is known'
- Figure of speech 'what `?` sees'

````rust
    /// The function is `async` and the future's output could not be
    /// projected, so nothing is known about what `?` sees
````

</details>

File: crates/whisker-rust/src/decorations/type_path.rs [hunk 10]
```diff
@@ -0,0 +1,213 @@
+use std::fmt;
+
+use crate::decorations::TypePathRef;
+
+/// Where a type is defined, as a path the compiler recognizes
```
martijnarts: Rewritten. Rules broken:

- Essay on rejected design alternatives (trait check, .context() check) is unrelated to the item
- Filler 'it is worth saying why'; near-aside 'Nominal identity is the honest primitive'
- 40-word sentence ('Rust-analyzer renders ... leaves out'); 35-word sentence ('Asking whether .context() can be called ...')
- Figures of speech 'leaves no trace', 'honest primitive', 'tempting alternatives'
- Speculative future-work sentence ('A rule that genuinely needs the public path ... nothing needs that today')

````rust
/// Where a type is defined, as a path the compiler would recognize
///
/// A display string cannot answer "is this `anyhow::Error`". Rust-analyzer
/// renders `syn::Error`, `std::io::Error`, `std::fmt::Error`, and
/// `anyhow::Error` all as `Error`, because it renders types the way a
/// programmer would write them at a use site, where the surrounding imports
/// supply what the text leaves out. A lint has no use site and no imports,
/// so it needs the identity the renderer discarded.
///
/// Two tempting alternatives do not work, and it is worth saying why.
/// Asking whether the type implements [`Error`] discriminates nothing:
/// `syn::Error`, `serde_json::Error`, `std::io::Error`, and everything
/// `anyhow::Error` wraps all implement it. Asking whether `.context()` can
/// be called on it also fails, because `anyhow::Context` is implemented for
/// [`Result<T, E>`] for every `E: Error + Send + Sync + 'static`, so the
/// method resolves perfectly well inside a `syn::Result` function. What
/// actually stops you writing `.context(..)?` there is the return
/// conversion, not method applicability. Nominal identity is the honest
/// primitive.
///
/// The path names the item's *definition*, not any public re-export of it.
/// `syn::Error` reads back as `syn::error::Error`, because that is where
/// `syn` declares it; the root re-export leaves no trace here. Comparing
/// definition paths is what makes the comparison stable: a crate may add,
/// move, or drop a re-export in a patch release, but moving a definition is
/// a breaking change. A rule that genuinely needs the public path instead
/// would have to resolve it through rust-analyzer's module scope, which
/// follows `pub use`; nothing needs that today.
///
/// # Examples
///
/// ```
/// use whisker_rust::decorations::TypePath;
///
/// let path = TypePath::new("syn", ["error"], "Error");
///
/// assert_eq!(path.to_string(), "syn::error::Error");
/// ```
///
/// [`Error`]: std::error::Error
/// [`Result<T, E>`]: std::result::Result
````

</details>

File: crates/whisker-rust/src/decorations/type_path.rs [hunk 11]
```diff
@@ -0,0 +1,213 @@
...
+    /// let path = TypePath::new("syn", ["error"], "Error");
+    ///
+    /// assert_eq!(path.name(), "Error");
+    /// ```
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+}
+
+/// Renders the path as `krate::module::Name`
```
martijnarts: Rewritten. Rules broken:

- Figure of speech 'cannot be fooled'; passive

````rust
/// Renders the path as `krate::module::Name`, for diagnostics only
///
/// Comparison goes through [`TypePathRef`], which cannot be fooled by a path
/// that happens to render the same way.
///
/// [`TypePathRef`]: crate::decorations::TypePathRef
````

</details>

File: crates/whisker-rust/src/decorations/type_path.rs [hunk 12]
```diff
@@ -0,0 +1,213 @@
...
+    #[test]
+    fn eq_with_different_crate_is_false() {
+        let path = TypePath::new("thiserror", [] as [&str; 0], "Error");
+
+        let matches = path == TypePathRef::new("anyhow", &[], "Error");
+
+        assert!(!matches);
+    }
+
+    /// Pins the explicit length check in the module comparison
```
martijnarts, line 162: Rewritten. Rules broken:

- Idioms 'pass for', 'answer to'; '-ing' opener 'Zipping'
- Test docs must be short and parseable

````rust
    /// A prefix of the right modules must not pass for the right modules
    ///
    /// Zipping two segment lists stops at the shorter one, so without an
    /// explicit length check `core::io::Error` would answer to a path naming
    /// `core::io::error::Error` and vice versa.
````

</details>

File: crates/whisker-rust/src/decorations/type_path_ref.rs [hunk 13]
```diff
@@ -0,0 +1,123 @@
+/// A borrowed [`TypePath`] a rule can declare as a `const`
```
martijnarts, line 1: Rewritten. Rules broken:

- Verbless fragment 'Const-constructible so ...'; 33-word sentence
- Rather-than (twice: 'instead of open-coding', 'rather than `TypeRef`')
- Naming-rationale essay unrelated to what the item does

````rust
/// A [`TypePath`] a lint can write down as a constant
///
/// Const-constructible so a rule states its subject once, at module scope,
/// beside the rule that cares about it, instead of open-coding a string
/// comparison at the point of use. Named `TypePathRef` rather than `TypeRef`
/// because `ra_ap_hir` re-exports a `TypeRef` of its own and the provider
/// imports from both.
///
/// # Examples
///
/// ```
/// use whisker_rust::decorations::{TypePath, TypePathRef};
///
/// const ANYHOW_ERROR: TypePathRef<'static> = TypePathRef::new("anyhow", &[], "Error");
///
/// assert!(TypePath::new("anyhow", [] as [&str; 0], "Error") == ANYHOW_ERROR);
/// assert!(TypePath::new("syn", ["error"], "Error") != ANYHOW_ERROR);
/// ```
///
/// [`TypePath`]: crate::decorations::TypePath
````

</details>

**Review by martijnarts — COMMENTED** [review]

File: crates/whisker-rust/src/decorations/type_path_ref.rs [hunk 14]
```diff
@@ -0,0 +1,126 @@
+/// A [`TypePath`] a lint can write down as a constant
+///
+/// Const-constructible so a rule states its subject once, at module scope,
+/// beside the rule that cares about it, instead of open-coding a string
+/// comparison at the point of use. Named `TypePathRef` rather than `TypeRef`
+/// because `ra_ap_hir` re-exports a `TypeRef` of its own and the provider
+/// imports from both.
```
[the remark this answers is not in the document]
martijnarts: You are right that the claim was false, and the sentence is already gone. The
naming note lived on an earlier push of this branch. `git grep TypeRef` over
`crates/` and `lints/` finds nothing, so no code here imports
`ra_ap_hir::TypeRef`.

I did not restore a corrected version. A paragraph that explains why the type
is not called something else answers a question no reader asks. It also goes
stale as soon as the import list changes.

The doc now says why the type exists at all: `TypePath` owns its segments and
cannot be a `const`, so `TypePathRef` borrows instead.

## Wider sweep — doc claims about other code
VERDICT: fixed
I checked every `///` line this branch adds against the code. Three claims did
not hold.

`TypePath`'s summary said the path is one "the compiler recognizes". It is
not. The next paragraph says the path names the definition, and `syn::error`
is a private module, so `syn::error::Error` does not resolve in user code. The
summary now says the path runs from the crate root.

`TypePath`'s doc also called definition paths stable, "because a re-export can
change in a patch release while a moved definition is a breaking change". That
is backwards. A crate can move a definition between private modules in a patch
release and keep the public re-export. So the definition path is the less
stable of the two. I dropped the justification and kept the invariant a caller
needs.

`TypePathRef`'s doc said a rule "compares it against a `TypePath` with `==`".
Only `PartialEq<TypePathRef> for TypePath` exists, so the other direction does
not compile. The doc now describes the type instead of the call.

Two smaller edits. `ErrorType` called a `TypePath` an "identity" where the
rest of the file says "path", so it now says "path" too. A test doc used
"would compare equal", where simple present reads better.

I checked the rest and left it alone. `ReturnMode`'s variant docs describe the
three cases the enum names, and nothing here contradicts them. The
`Box<dyn Error>` sentence in `ErrorType` follows from the variant's own
definition, because `Box` is an ADT.

One note on verification. In a private target directory, `cargo fmt --check`,
`cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test
--all-features` all pass, and the eleven doctests in `whisker-rust` are among
them.

PR #218, opened 2026-08-05
Resolve error types by identity, not by rendering

martijnarts:
`anyhow_missing_context` exists to catch `foo()?` used without `.context(..)`
in a function returning `anyhow::Result`. Running it over this repository
produced ten diagnostics, seven of them wrong: four in `whisker-macros`, which
does not depend on anyhow at all, two on `fn fmt(..) -> fmt::Result`, where
`.context()` is not even expressible, and one on a `serde_json::Error`. The
rule's own documentation promises it fires only for `anyhow::Error`.

The provider now resolves a function's error type to the item that defines it
and hands the rule a canonical path, which the rule compares against a
constant. Whether a type is a `Result` or an `Option` goes through the same
route, not through its rendering.

One more blind spot went with it. `Function::ret_type` gives an `async fn` the
opaque future, not the awaited type, so the error type was absent for every
async function. Every whisker CLI command is async, so the rule could not see
any of them. `Function::async_ret_type` exists for exactly this case and is
now used.

The old tests never noticed any of this because they hand-build their
decorations, so the rule was only ever exercised against input the real
provider would never produce. The tests here run the actual provider over a
fixture holding `anyhow::Result`, `syn::Result`, `io::Result`, `fmt::Result`,
a boxed error, a generic error, async functions, and a closure returning a
different error type than the function around it. They assert against the
whole file, not one function at a time, because a per-function expectation
cannot fail on a rule that starts firing somewhere else.

### DIFF
--- a/crates/whisker-rust/src/decorations/fn_signature.rs
+++ b/crates/whisker-rust/src/decorations/fn_signature.rs
@@ -1,57 +1,80 @@
-use crate::decorations::ResolvedType;
+use crate::decorations::{ErrorType, ResolvedType, ReturnMode};
 
 /// Function signature information
 ///
-/// Attached to function item nodes so lints can inspect the return type
-/// without re-querying the semantic model.
+/// Attached to `function_item` nodes so lints can inspect the return type
+/// without re-querying the semantic model. Every resolved function gets a
+/// signature, even an empty one. An absent decoration would look like a
+/// function the provider never reached, and those are different facts.
 #[derive(Clone, Eq, PartialEq, Hash, Debug)]
 pub struct FnSignature {
     return_type: Option<ResolvedType>,
-    error_type_name: Option<String>,
+    error_type: Option<ErrorType>,
+    return_mode: ReturnMode,
 }
 
 impl FnSignature {
     /// Creates a function signature
-    pub fn new(return_type: Option<ResolvedType>, error_type_name: Option<String>) -> Self {
+    pub fn new(
+        return_type: Option<ResolvedType>,
+        error_type: Option<ErrorType>,
+        return_mode: ReturnMode,
+    ) -> Self {
         Self {
             return_type,
-            error_type_name,
+            error_type,
+            return_mode,
         }
     }
 
-    /// Returns the resolved return type, if available
+    /// Returns the resolved return type, awaited for an `async fn`
     pub fn return_type(&self) -> Option<&ResolvedType> {
         self.return_type.as_ref()
     }
 
-    /// Returns the error type name for `Result`-returning functions
+    /// Returns the `E` of the `Result<T, E>` this function returns
     ///
-    /// This is the fully qualified name of the `E` in `Result<T, E>`,
-    /// e.g. `"anyhow::Error"`.
-    pub fn error_type_name(&self) -> Option<&str> {
-        self.error_type_name.as_deref()
+    /// [`None`] means the function does not return a `core::result::Result`.
+    /// That includes an `async fn` whose awaited type is not one, and one
+    /// whose future could not be projected.
+    pub fn error_type(&self) -> Option<&ErrorType> {
+        self.error_type.as_ref()
+    }
+
+    /// Returns where the return type was read from
+    pub fn return_mode(&self) -> ReturnMode {
+        self.return_mode
     }
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::decorations::{TypePath, TypePathRef};
 
     #[test]
     fn fn_signature_with_return_type() {
         let ret = ResolvedType::new("Result<(), anyhow::Error>".into()).with_result(true);
-        let sig = FnSignature::new(Some(ret), Some("anyhow::Error".into()));
+        let error = ErrorType::Named(TypePath::new("anyhow", [] as [&str; 0], "Error"));
+
+        let sig = FnSignature::new(Some(ret), Some(error), ReturnMode::Direct);
 
         assert!(sig.return_type().unwrap().is_result());
-        assert_eq!(sig.error_type_name(), Some("anyhow::Error"));
+        assert!(
+            sig.error_type()
+                .unwrap()
+                .is(TypePathRef::new("anyhow", &[], "Error"))
+        );
+        assert_eq!(sig.return_mode(), ReturnMode::Direct);
     }
 
     #[test]
     fn fn_signature_without_return_type() {
-        let sig = FnSignature::new(None, None);
+        let sig = FnSignature::new(None, None, ReturnMode::Opaque);
 
         assert!(sig.return_type().is_none());
-        assert!(sig.error_type_name().is_none());
+        assert!(sig.error_type().is_none());
+        assert_eq!(sig.return_mode(), ReturnMode::Opaque);
     }
 
     #[test]
--- a/crates/whisker-rust/src/provider.rs
+++ b/crates/whisker-rust/src/provider.rs
@@ -1,20 +1,22 @@
+use std::ops::Range;
 use std::path::Path;
 use std::sync::{Arc, Mutex};
 
 use anyhow::Context as _;
-use ra_ap_hir::{Adt, HasAttrs, HirDisplay, Semantics, attach_db};
+use ra_ap_hir::{Adt, Enum, Function, HasAttrs, HirDisplay, Module, Semantics, Type, attach_db};
 use ra_ap_ide_db::base_db::SourceDatabase as _;
+use ra_ap_ide_db::famous_defs::FamousDefs;
 use ra_ap_ide_db::{ChangeWithProcMacros, RootDatabase};
 use ra_ap_load_cargo::{LoadCargoConfig, ProcMacroServerChoice};
 use ra_ap_project_model::{CargoConfig, ProjectManifest, ProjectWorkspace};
-use ra_ap_syntax::{AstNode, NodeOrToken, SyntaxNode, TextRange, TextSize, ast};
+use ra_ap_syntax::{AstNode, Edition, NodeOrToken, SyntaxNode, TextRange, TextSize, ast};
 use ra_ap_vfs::{AbsPathBuf, FileExcluded, FileId, Vfs, VfsPath};
 use whisker_types::{
     Coverage, CoverageGap, DecoratedNode, DecoratedTree, DecorationMap, DecorationProvider,
     ProviderName,
 };
 
-use crate::decorations::{AdtFlags, FnSignature, ResolvedType};
+use crate::decorations::{AdtFlags, ErrorType, FnSignature, ResolvedType, ReturnMode, TypePath};
 
 /// Decoration provider for Rust using rust-analyzer
 ///
@@ -262,18 +264,18 @@ impl DecorationProvider for RustDecorationProvider {
         let coverage = attach_db(&*db, || {
             let sema = Semantics::new(&*db);
 
-            if sema.file_to_module_def(file_id).is_none() {
+            let Some(module) = sema.file_to_module_def(file_id) else {
                 return Coverage::NotCovered(CoverageGap::Unreachable {
                     root: Arc::clone(&self.root),
                 });
-            }
+            };
 
             let db_text: &str = db.file_text(file_id).text(&*db);
             if db_text != tree.source() {
                 return Coverage::NotCovered(CoverageGap::StaleSource);
             }
 
-            Coverage::Covered(resolve_targets(&sema, file_id, &targets))
+            Coverage::Covered(resolve_targets(&sema, file_id, module, &targets))
         });
 
         Ok(coverage)
@@ -282,19 +284,30 @@ impl DecorationProvider for RustDecorationProvider {
 
 /// Resolves every collected target against rust-analyzer's parse
 ///
-/// The caller has already checked that `file_id` belongs to a module and
-/// that its recorded text matches the source the targets index into.
+/// The caller has already established that `file_id` belongs to `module`
+/// and that its recorded text matches the text the targets' byte ranges
+/// were taken from.
+///
+/// `module` identifies the crate. The crate supplies the edition for
+/// display and the `core` definitions of `Result` and `Option`. These
+/// resolve once per file, not once per target.
 fn resolve_targets(
     sema: &Semantics<'_, RootDatabase>,
     file_id: FileId,
+    module: Module,
     targets: &[Target],
 ) -> DecorationMap {
     let source_file = sema.parse_guess_edition(file_id);
     let syntax = source_file.syntax().clone();
 
+    let krate = module.krate(sema.db);
+    let famous = FamousDefs(sema, krate);
     let ctx = DecorateCtx {
         sema,
         syntax: &syntax,
+        edition: krate.edition(sema.db),
+        core_result: famous.core_result_Result(),
+        core_option: famous.core_option_Option(),
     };
 
     let mut decorations = DecorationMap::new();
@@ -367,17 +380,25 @@ enum Target {
     },
 }
 
-/// Converts a tree-sitter byte range into a rust-analyzer [`TextRange`]
+/// Converts a byte range into rust-analyzer's 32-bit one
 ///
-/// Rust-analyzer stores text offsets as 32 bits. When an offset does not
-/// fit, the function returns [`None`]. The caller then skips the target,
-/// because a truncated offset would name some other expression.
-fn text_range_of(node: &DecoratedNode<'_>) -> Option<TextRange> {
-    let range = node.raw().byte_range();
-    let start = u32::try_from(range.start).ok()?;
-    let end = u32::try_from(range.end).ok()?;
+/// Rust-analyzer addresses text with 32-bit offsets. An offset that does
+/// not fit yields [`None`], because a truncated offset would name some
+/// other expression.
+///
+/// The conversion is a separate function so that a unit test can reach the
+/// 32-bit boundary without a four-gigabyte file.
+fn text_range_from(range: Range<usize>) -> Option<TextRange> {
+    let Range { start, end } = range;
+    let start = u32::try_from(start).ok()?;
+    let end = u32::try_from(end).ok()?;
 
     Some(TextRange::new(TextSize::from(start), TextSize::from(end)))
+}
+
+/// Converts a tree-sitter node's byte range into rust-analyzer's
+fn text_range_of(node: &DecoratedNode<'_>) -> Option<TextRange> {
+    text_range_from(node.raw().byte_range())
 }
 
 fn collect_targets(node: &DecoratedNode<'_>, targets: &mut Vec<Target>) {
@@ -429,13 +450,20 @@ fn collect_targets(node: &DecoratedNode<'_>, targets: &mut Vec<Target>) {
     }
 }
 
+/// Shared state for resolving one file's targets
+///
+/// The edition and the `core` enums are per-crate, so `resolve_targets`
+/// fills them once for the whole file.
 struct DecorateCtx<'a, 'db> {
     sema: &'a Semantics<'db, RootDatabase>,
-    syntax: &'a ra_ap_syntax::SyntaxNode,
+    syntax: &'a SyntaxNode,
+    edition: Edition,
+    core_result: Option<Enum>,
+    core_option: Option<Enum>,
 }
 
 impl DecorateCtx<'_, '_> {
-    fn display_type(&self, ty: &ra_ap_hir::Type<'_>, func: &ra_ap_hir::Function) -> String {
+    fn display_type(&self, ty: &Type<'_>, func: &Function) -> String {
         let krate = func.module(self.sema.db).krate(self.sema.db);
         let target = krate.to_display_target(self.sema.db);
         format!("{}", ty.display(self.sema.db, target))
@@ -444,57 +472,152 @@ impl DecorateCtx<'_, '_> {
 
 /// Returns the smallest node in rust-analyzer's tree that covers `range`
 ///
-/// Tree-sitter's parse and rust-analyzer's parse agree on byte offsets but
-/// not on tree shape, so the function locates a node by its span. Callers
-/// climb from the covering node to the smallest [`ast::Expr`] or
-/// [`ast::Fn`] that encloses the whole span.
+/// The two parsers agree on byte offsets but not on tree shape, so this
+/// function locates a tree-sitter node by the span it occupies. From the
+/// smallest covering node, the callers climb to the smallest [`ast::Expr`]
+/// or [`ast::Fn`] that encloses the whole span.
 ///
-/// The function returns [`None`] for an empty range or a range outside the
-/// tree. Tree-sitter synthesizes zero-width nodes during error recovery,
-/// and a zero-width span "covers" some unrelated neighbor.
-fn covering_node(ctx: &DecorateCtx<'_, '_>, range: TextRange) -> Option<SyntaxNode> {
-    if range.is_empty() || !ctx.syntax.text_range().contains_range(range) {
+/// The function refuses an empty range. Tree-sitter synthesizes zero-width
+/// nodes during error recovery, and a zero-width span sits inside some
+/// token, so the broken node would take that token's type.
+///
+/// It also refuses a range past the end of `syntax`.
+/// [`SyntaxNode::covering_element`] panics on such a range, and one bad
+/// target would abort the whole run.
+///
+/// The function takes the tree instead of the whole [`DecorateCtx`] so
+/// that unit tests can call it without a loaded workspace.
+fn covering_node(syntax: &SyntaxNode, range: TextRange) -> Option<SyntaxNode> {
+    if range.is_empty() || !syntax.text_range().contains_range(range) {
         return None;
     }
 
-    match ctx.syntax.covering_element(range) {
+    match syntax.covering_element(range) {
         NodeOrToken::Node(node) => Some(node),
         NodeOrToken::Token(token) => token.parent(),
     }
 }
 
-fn find_enclosing_fn(ctx: &DecorateCtx<'_, '_>, range: TextRange) -> Option<ra_ap_hir::Function> {
-    let fn_node = covering_node(ctx, range)?
+fn find_enclosing_fn(ctx: &DecorateCtx<'_, '_>, range: TextRange) -> Option<Function> {
+    let fn_node = covering_node(ctx.syntax, range)?
         .ancestors()
         .find_map(ast::Fn::cast)?;
     ctx.sema.to_def(&fn_node)
 }
 
 fn find_expr_covering(ctx: &DecorateCtx<'_, '_>, range: TextRange) -> Option<ast::Expr> {
-    covering_node(ctx, range)?
+    covering_node(ctx.syntax, range)?
         .ancestors()
         .find_map(ast::Expr::cast)
+}
+
+/// Returns the return type as seen from inside the function's body
+///
+/// For an `async fn`, that is the future's output. In the body, `?` and
+/// `return` convert into the output, not into the opaque `impl Future`.
+///
+/// The projection uses [`Function::async_ret_type`], which returns a type
+/// only for an `async fn`. A `Future`-implementation test would also match
+/// `fn f() -> impl Future<Output = anyhow::Result<()>>`, where the body's
+/// `?` does not target that [`Result<T, E>`].
+///
+/// [`Result<T, E>`]: std::result::Result
+fn effective_return_type<'db>(
+    ctx: &DecorateCtx<'_, 'db>,
+    func: Function,
+) -> (Option<Type<'db>>, ReturnMode) {
+    if func.is_async(ctx.sema.db) {
+        match func.async_ret_type(ctx.sema.db) {
+            Some(ty) => (Some(ty), ReturnMode::Awaited),
+            None => (None, ReturnMode::Opaque),
+        }
+    } else {
+        (Some(func.ret_type(ctx.sema.db)), ReturnMode::Direct)
+    }
+}
+
+/// Returns whether `ty` is an instance of the enum `target` names
+///
+/// The comparison is by identity, not by rendered name: a user's own
+/// `myres::Result<T>` renders as `Result<..>` but does not match.
+fn is_instance_of(ty: &Type<'_>, target: Option<Enum>) -> bool {
+    let Some(target) = target else {
+        return false;
+    };
+
+    match ty.as_adt() {
+        Some(Adt::Enum(actual)) => actual == target,
+        Some(Adt::Struct(_)) => false,
+        Some(Adt::Union(_)) => false,
+        None => false,
+    }
+}
+
+/// Returns the path of `ty`'s definition, if it has one
+///
+/// The path is the definition's, not a re-export's.
+fn type_path(ctx: &DecorateCtx<'_, '_>, ty: &Type<'_>) -> Option<TypePath> {
+    let db = ctx.sema.db;
+    let adt = ty.as_adt()?;
+    let module = adt.module(db);
+    let krate = module.krate(db).display_name(db)?;
+    let modules: Vec<String> = module
+        .path_segments(db)
+        .map(|segment| segment.display(db, ctx.edition).to_string())
+        .collect();
+    let name = adt.name(db).display(db, ctx.edition).to_string();
+
+    Some(TypePath::new(&krate.to_string(), modules, &name))
+}
+
+/// Classifies the `E` of a [`Result<T, E>`]
+///
+/// A type parameter becomes [`ErrorType::Generic`]; a type with no
+/// definition path at all becomes [`ErrorType::Unnamed`]. Neither gets an
+/// invented name, because callers compare the result against real paths.
+///
+/// [`Result<T, E>`]: std::result::Result
+fn classify_error(ctx: &DecorateCtx<'_, '_>, ty: &Type<'_>) -> ErrorType {
+    match type_path(ctx, ty) {
+        Some(path) => ErrorType::Named(path),
+        None => match ty.as_type_param(ctx.sema.db) {
+            Some(_) => ErrorType::Generic,
+            None => ErrorType::Unnamed,
+        },
+    }
+}
+
+/// Returns the classified `E` when `ty` really is `core::result::Result<T, E>`
+fn result_error_type(ctx: &DecorateCtx<'_, '_>, ty: &Type<'_>) -> Option<ErrorType> {
+    let core_result = ctx.core_result?;
+    let (adt, args) = ty.as_adt_with_args()?;
+    let Adt::Enum(actual) = adt else {
+        return None;
+    };
+    if actual != core_result {
+        return None;
+    }
+
+    let error = args.into_iter().nth(1)??;
+    Some(classify_error(ctx, &error))
 }
 
 fn resolve_function(ctx: &DecorateCtx<'_, '_>, range: TextRange) -> Option<FnSignature> {
     let func = find_enclosing_fn(ctx, range)?;
 
-    let ret_type = func.ret_type(ctx.sema.db);
-    let display = ctx.display_type(&ret_type, &func);
-    let is_result = display.starts_with("Result<") || display.contains("::Result<");
-    let is_never = ret_type.is_never();
-
-    let resolved = ResolvedType::new(display.clone())
-        .with_result(is_result)
-        .with_never(is_never);
-
-    let error_type_name = if is_result {
-        extract_result_error_type(&display)
-    } else {
-        None
+    let (ret_type, return_mode) = effective_return_type(ctx, func);
+    let Some(ret_type) = ret_type else {
+        return Some(FnSignature::new(None, None, return_mode));
     };
 
-    Some(FnSignature::new(Some(resolved), error_type_name))
+    let display = ctx.display_type(&ret_type, &func);
+    let error_type = result_error_type(ctx, &ret_type);
+
+    let resolved = ResolvedType::new(display)
+        .with_result(is_instance_of(&ret_type, ctx.core_result))
+        .with_never(ret_type.is_never());
+
+    Some(FnSignature::new(Some(resolved), error_type, return_mode))
 }
 
 fn resolve_match_scrutinee(
@@ -539,8 +662,8 @@ fn resolve_expr_type(ctx: &DecorateCtx<'_, '_>, range: TextRange) -> Option<Reso
         None => format!("{ty:?}"),
     };
 
-    let is_result = display.starts_with("Result<") || display.contains("::Result<");
-    let is_option = display.starts_with("Option<") || display.contains("::Option<");
+    let is_result = is_instance_of(&ty, ctx.core_result);
+    let is_option = is_instance_of(&ty, ctx.core_option);
     let is_never = ty.is_never();
 
     Some(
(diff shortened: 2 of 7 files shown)


martijnarts:
Every doc comment this PR adds was re-checked against three rule sets and rewritten where it failed. 32 blocks changed here.

Violations found here: 1 stated something untrue at this branch, 17 reviewer verdicts from #208, 40 ASD-STE100, 25 AI-writing tells.

|---|---|---|
| `crates/whisker-rust/src/decorations/fn_signature.rs` | fn FnSignature::error_type | 29-word sentence |
| `crates/whisker-rust/src/decorations/fn_signature.rs` | struct FnSignature | 33-word sentence; Passive 'could be established', 'is attached' |
| `crates/whisker-rust/src/provider.rs` | const COVERING_SOURCE (test fixture) | Test-helper doc too long, 'unparseable' second paragraph; 'That is what makes ... observable' construction; Passive 'nothing is returned' |
| `crates/whisker-rust/src/provider.rs` | fn classify_error | 33-word sentence; Figurative 'make that comparison lie'; Does not say what the classifications are |
| `crates/whisker-rust/src/provider.rs` | fn covering_node | 45-word sentence in the shape paragraph; 26-word colon sentence in the added past-the-end paragraph; Parameter-choice essay ('a parameter that needs a loaded workspace to construct is one... |
| `crates/whisker-rust/src/provider.rs` | fn effective_return_type | Riddle summary 'the type `?` and `return` answer to'; 'answer confidently and wrongly', significance inflation 'exactly the distinction that matters'; 34-word sentence; Verbless fragment ... |
| `crates/whisker-rust/src/provider.rs` | fn is_instance_of | Verbless fragment 'Identity, not rendering:'; Vague 'and is not this' |
| `crates/whisker-rust/src/provider.rs` | fn resolve_targets | Implementation-detaily essay about crate fixing; 36-word sentence with figurative 'is what fixes the crate every question ... is asked on behalf of'; -ing clause 'resolving them once per ... |
| `crates/whisker-rust/src/provider.rs` | fn text_range_from | Figurative 'the only honest answer', 'quietly name'; 'decorate this node' but the function takes a range, not a node; Test-design essay ('Fused to a DecoratedNode ... four-gigabyte fixtur... |
| `crates/whisker-rust/src/provider.rs` | fn type_path | 28-word sentence |
| `crates/whisker-rust/src/provider.rs` | struct DecorateCtx | Summary is a riddle, does not say what the item is; Figurative 'the whole difference between asking ... what a type *is* and reading what it *printed*'; No finite structure, italic emphasis |
| `crates/whisker-rust/src/provider.rs` | test covering_node_with_empty_range_returns_none | Test doc over two sentences; 33-word sentence; Figurative 'borrow the identity' |
| `crates/whisker-rust/src/provider.rs` | test covering_node_with_range_past_the_end_returns_none | -ing cluster 'the difference between skipping ... and aborting'; Passive summary 'must be declined, not asserted on' |
| `crates/whisker-rust/src/provider.rs` | test text_range_from_with_offset_past_u32_returns_none | Rather-than ('confidently wrong rather than absent'); -ing clauses 'Truncating', 'Reaching'; Test doc wordier than the pin needs |
| `crates/whisker-rust/tests/provider_integration.rs` | fn count_fn_signatures | Em-dash aside pair '— the method call behind `?`, the `async` trait method —'; 41-word sentence; Inflation 'exactly the ones' |
| `crates/whisker-rust/tests/provider_integration.rs` | fn flagged_in_file | 27-word convoluted sentence 'Listing the whole file is the expectation that has to be updated ...'; -ing openers 'Listing', 'firing'; 'which is what makes ... visible' construction |
| `crates/whisker-rust/tests/provider_integration.rs` | test anyhow_missing_context_flags_a_bare_try_in_an_async_anyhow_function | CLI-history aside is unrelated to what the test pins; Idiom 'the rule gives up'; -ing clauses 'returning', 'reading' |
| `crates/whisker-rust/tests/provider_integration.rs` | test anyhow_missing_context_flags_a_bare_try_in_an_async_trait_method | -ing clause 'declares the method as returning' |
| `crates/whisker-rust/tests/provider_integration.rs` | test anyhow_missing_context_flags_only_the_anyhow_function_among_lookalike_errors | Unparseable essay, five sentences plus a second paragraph; 30-word sentence 'This runs the real provider ... happen to agree with it'; Figurative 'the sharp one' |
| `crates/whisker-rust/tests/provider_integration.rs` | test anyhow_missing_context_ignores_a_bare_try_in_an_async_io_function | Figurative 'trade one wrong answer for another'; -ing opener 'Reading the awaited type', 33-word sentence |
| `crates/whisker-rust/tests/provider_integration.rs` | test anyhow_missing_context_ignores_a_bare_try_in_an_io_function | Idiom 'not this rule's business'; 'pins the identity question against the real provider' sentence duplicates the lookalike test's rationale; -ing clauses 'returning', 'reading' |
| `crates/whisker-rust/tests/provider_integration.rs` | test anyhow_missing_context_ignores_a_boxed_error | Filler 'simply', idiom 'the one it cares about' |
| `crates/whisker-rust/tests/provider_integration.rs` | test anyhow_missing_context_ignores_a_generic_error | 31-word sentence; Filler 'may well', idiom 'stay quiet' |
| `crates/whisker-rust/tests/provider_integration.rs` | test anyhow_missing_context_ignores_a_local_error_type | Idiom 'answer to'; 40-word sentence with trailing appositive; History detail ('four of the seven') beyond what the pin needs |
| `crates/whisker-rust/tests/provider_integration.rs` | test anyhow_missing_context_ignores_a_try_inside_a_closure | -ing cluster 'removing the barrier fails it by gaining an entry instead of losing one'; Last sentence unparseable |
| `crates/whisker-rust/tests/provider_integration.rs` | test anyhow_missing_context_over_the_whole_file_reports_only_anyhow_bodies | Near-duplicates flagged_in_file's rationale sentence; 33-word sentence; Test doc over two sentences |
| `crates/whisker-rust/tests/provider_integration.rs` | test fn_signature_on_async_function_reports_the_awaited_type | Test doc over two sentences; Figurative 'fail loudly'; 29-word sentence |
| `lints/anyhow_missing_context/src/lib.rs` | const ANYHOW_ERROR | Summary is a riddle, does not name what the constant holds; 25-word sentence with awkward 'can never match this however the type prints' |
| `lints/anyhow_missing_context/src/lib.rs` | const BODY_BARRIERS | Em-dash aside '— and, worse, —'; Rather-than ('rather than the enclosing function's'); 33-word sentence |
| `lints/anyhow_missing_context/src/lib.rs` | fn find_enclosing_fn_signature | Em-dash aside pair around the false-negative clause; -ing clauses 'on reaching ... before reaching'; Hazy 'the right direction for a rule that runs in CI' |
| `lints/anyhow_missing_context/src/lib.rs` | struct AnyhowMissingContext | Significance inflation 'significantly easier'; 32-word sentence in the added paragraph; -ing opener 'Suggesting .context(..) there' |
| `lints/anyhow_missing_context/src/lib.rs` | test in_closure_inside_anyhow_fn_is_not_flagged | -ing opener 'Reporting', -ing clause 'closure returning something else'; Near-duplicate of the compile rationale already in the rule doc |

- `crates/whisker-rust/src/provider.rs` &mdash; fn text_range_from: 'decorate this node' but the function takes a range, not a node

martijnarts (commented):

crates/whisker-rust/src/provider.rs
```diff
@@ -282,19 +284,30 @@ impl DecorationProvider for RustDecorationProvider {
 
 /// Resolves every collected target against rust-analyzer's parse
 ///
-/// The caller has already checked that `file_id` belongs to a module and
-/// that its recorded text matches the source the targets index into.
+/// The caller has already established that `file_id` belongs to `module`
```
martijnarts, line 287: Rewritten. Rules broken:

- Implementation-detaily essay about crate fixing
- 36-word sentence with figurative 'is what fixes the crate every question ... is asked on behalf of'
- -ing clause 'resolving them once per file'

````rust
/// Resolves every collected target against rust-analyzer's parse
///
/// The caller has already established that `file_id` belongs to `module`
/// and that its recorded text matches the text the targets' byte ranges
/// were taken from.
///
/// The module is what fixes the crate every question in this file is asked
/// on behalf of: the edition its names print in, and the `core` whose
/// `Result` and `Option` a type is compared against. Every function in one
/// file belongs to one crate, so resolving them once per file costs less
/// than once per target and loses nothing.
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 2]
```diff
@@ -367,19 +380,27 @@ enum Target {
     },
 }
 
-/// Converts a tree-sitter byte range into a rust-analyzer [`TextRange`]
+/// Converts a byte range into rust-analyzer's 32-bit one
```
martijnarts, line 383: Rewritten. Rules broken:

- Figurative 'the only honest answer', 'quietly name'
- 'decorate this node' but the function takes a range, not a node
- Test-design essay ('Fused to a DecoratedNode ... four-gigabyte fixture') longer than the item warrants
- 27-word sentence

````rust
/// Converts a byte range into rust-analyzer's 32-bit one
///
/// Rust-analyzer addresses text with 32-bit offsets, so a file too large for
/// one is a file it cannot describe. Returning [`None`] is then the only
/// honest answer, because a truncated offset would quietly name some other
/// expression and decorate this node with that expression's type.
///
/// The arithmetic is separate from the node access so that the refusal can be
/// tested. Fused to a [`DecoratedNode`], the only way to reach the boundary
/// would be a four-gigabyte fixture.
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 3]
```diff
@@ -429,13 +450,20 @@ fn collect_targets(node: &DecoratedNode<'_>, targets: &mut Vec<Target>) {
     }
 }
 
+/// Shared state for resolving one file's targets
```
martijnarts, line 453: Rewritten. Rules broken:

- Summary is a riddle, does not say what the item is
- Figurative 'the whole difference between asking ... what a type *is* and reading what it *printed*'
- No finite structure, italic emphasis

````rust
/// What every question about one file is asked against
///
/// The three resolved fields are per-crate, not per-target, resolved once
/// and reused, because they are the whole difference between asking
/// rust-analyzer what a type *is* and reading what it *printed*.
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 4]
```diff
@@ -444,57 +472,154 @@ impl DecorateCtx<'_, '_> {
 
 /// Returns the smallest node in rust-analyzer's tree that covers `range`
 ///
-/// Tree-sitter's parse and rust-analyzer's parse agree on byte offsets but
-/// not on tree shape, so the function locates a node by its span. Callers
-/// climb from the covering node to the smallest [`ast::Expr`] or
-/// [`ast::Fn`] that encloses the whole span.
+/// The two parsers agree on byte offsets but not on tree shape, so this
```
martijnarts, line 475: Rewritten. Rules broken:

- 45-word sentence in the shape paragraph
- 26-word colon sentence in the added past-the-end paragraph
- Parameter-choice essay ('a parameter that needs a loaded workspace to construct is one no unit test can reach')
- -ing clauses 'Descending', 'reaching'

````rust
/// Returns the smallest node in rust-analyzer's tree that covers `range`
///
/// The two parsers agree on byte offsets, because the caller has already
/// checked that they were handed the same text, but not on tree shape. A
/// tree-sitter node is therefore located by the span it occupies, not by
/// position in the tree. Descending to the smallest covering node and then
/// climbing is what makes the lookup shape-independent: whichever node
/// rust-analyzer happens to have at that span, the ancestor walk from it
/// reaches the smallest [`ast::Expr`] or [`ast::Fn`] enclosing the whole
/// span, which is the one the caller meant.
///
/// An empty range is refused. Tree-sitter synthesizes zero-width nodes while
/// recovering from a parse error, and a zero-width span "covers" whatever
/// token it sits inside, which would decorate the broken node with an
/// unrelated neighbor's type.
///
/// A range reaching past the end of `syntax` is refused too, and it has to be
/// refused here: [`SyntaxNode::covering_element`] asserts on one instead of
/// returning, and the caller would abort where it should skip a target.
///
/// The parameter is the tree, not the whole [`DecorateCtx`], because both
/// refusals are properties of a range against a parse, and a parameter that
/// needs a loaded workspace to construct is one no unit test can reach.
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 5]
```diff
@@ -444,57 +472,154 @@ impl DecorateCtx<'_, '_> {
...
 }
 
 fn find_expr_covering(ctx: &DecorateCtx<'_, '_>, range: TextRange) -> Option<ast::Expr> {
-    covering_node(ctx, range)?
+    covering_node(ctx.syntax, range)?
         .ancestors()
         .find_map(ast::Expr::cast)
 }
 
+/// Returns the return type as seen from inside the function's body
```
martijnarts, line 514: Rewritten. Rules broken:

- Riddle summary 'the type `?` and `return` answer to'
- 'answer confidently and wrongly', significance inflation 'exactly the distinction that matters'
- 34-word sentence
- Verbless fragment 'For an `async fn` that is the future's output, not the future:'

````rust
/// Returns the type `?` and `return` answer to inside the function's body
///
/// For an `async fn` that is the future's output, not the future: an
/// `async fn`'s signature output is the opaque `impl Future`, which is not
/// what any expression in its body converts into.
///
/// The projection deliberately goes through [`Function::async_ret_type`],
/// never a "does this implement `Future`" test. The latter also
/// sees through `fn f() -> impl Future<Output = anyhow::Result<()>>`, where a
/// `?` in the body cannot target that [`Result<T, E>`] at all, and would
/// answer confidently and wrongly. Gating on `is_async`, which
/// `async_ret_type` does itself, is exactly the distinction that matters.
///
/// [`Result<T, E>`]: std::result::Result
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 6]
```diff
@@ -444,57 +472,154 @@ impl DecorateCtx<'_, '_> {
...
+        match func.async_ret_type(ctx.sema.db) {
+            Some(ty) => (Some(ty), ReturnMode::Awaited),
+            None => (None, ReturnMode::Opaque),
+        }
+    } else {
+        (Some(func.ret_type(ctx.sema.db)), ReturnMode::Direct)
+    }
+}
+
+/// Returns whether `ty` is an instance of the enum `target` names
```
martijnarts, line 539: Rewritten. Rules broken:

- Verbless fragment 'Identity, not rendering:'
- Vague 'and is not this'

````rust
/// Returns whether `ty` is an instance of the enum `target` names
///
/// Identity, not rendering: a user's own `myres::Result<T>` renders as
/// `Result<..>` and is not this.
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 7]
```diff
@@ -444,57 +472,154 @@ impl DecorateCtx<'_, '_> {
...
+
+    match ty.as_adt() {
+        Some(Adt::Enum(actual)) => actual == target,
+        Some(Adt::Struct(_)) => false,
+        Some(Adt::Union(_)) => false,
+        None => false,
+    }
+}
+
+/// Returns the path of `ty`'s definition, if it has one
```
martijnarts, line 556: Rewritten. Rules broken:

- 28-word sentence

````rust
/// Returns where `ty`'s defining item lives, if it has one
///
/// The path is the definition's, not any re-export's, because a definition
/// only moves in a breaking release while a re-export can appear or vanish
/// in a patch one.
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 8]
```diff
@@ -444,57 +472,154 @@ impl DecorateCtx<'_, '_> {
...
+    let modules: Vec<String> = module
+        .path_segments(db)
+        .map(|segment| segment.display(db, ctx.edition).to_string())
+        .collect();
+    let name = adt.name(db).display(db, ctx.edition).to_string();
+
+    Some(TypePath::new(&krate.to_string(), modules, &name))
+}
+
+/// Classifies the `E` of a [`Result<T, E>`]
```
martijnarts, line 573: Rewritten. Rules broken:

- 33-word sentence
- Figurative 'make that comparison lie'
- Does not say what the classifications are

````rust
/// Classifies the `E` of a [`Result<T, E>`]
///
/// A type with no defining item is reported as such, never given an
/// invented name, because the only use of this value is a comparison against
/// a path, and a name nothing declares can only make that comparison lie.
///
/// [`Result<T, E>`]: std::result::Result
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 9]
```diff
@@ -551,31 +676,39 @@ fn resolve_expr_type(ctx: &DecorateCtx<'_, '_>, range: TextRange) -> Option<Reso
...
-
 #[cfg(test)]
 mod tests {
     use std::path::PathBuf;
 
+    use ra_ap_syntax::SourceFile;
+
     use super::*;
 
+    /// The source every `covering_node` test locates a range in
```
martijnarts, line 685: Rewritten. Rules broken:

- Test-helper doc too long, 'unparseable' second paragraph
- 'That is what makes ... observable' construction
- Passive 'nothing is returned'

````rust
    /// The source every `covering_node` test locates a range in
    ///
    /// `value` is more than one byte long, so an offset strictly inside it is
    /// available. That is what makes the empty-range refusal observable: at
    /// such an offset rowan does have an element to answer with, and the
    /// guard is the only reason nothing is returned.
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 10]
```diff
@@ -616,33 +749,45 @@ mod tests {
         }
     }
 
+    /// A zero-width span must not take the type of the token around it
```
martijnarts, line 750: Rewritten. Rules broken:

- Test doc over two sentences
- 33-word sentence
- Figurative 'borrow the identity'

````rust
    /// A zero-width span must not borrow the identity of the token around it
    ///
    /// Rowan answers `covering_element` for an empty range with the token it
    /// falls inside, so without the refusal the caller would decorate a
    /// tree-sitter node of no width with that token's type. The offset here
    /// sits in the middle of `value`, which is what makes the difference
    /// observable: the same offset with any width at all does resolve.
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 11]
```diff
@@ -616,33 +749,45 @@ mod tests {
...
+    fn covering_node_with_range_inside_the_file_returns_a_node() {
+        let syntax = covering_source_tree();
+        let value = offset_of("value");
+
+        let covering = covering_node(&syntax, range_at(value, 5)).expect("should cover `value`");
+
+        assert!(covering.text_range().contains_range(range_at(value, 5)));
     }
 
+    /// A span past the end of the tree must be skipped, not asserted on
```
martijnarts, line 777: Rewritten. Rules broken:

- -ing cluster 'the difference between skipping ... and aborting'
- Passive summary 'must be declined, not asserted on'

````rust
    /// A span the parse does not reach must be declined, not asserted on
    ///
    /// [`SyntaxNode::covering_element`] panics on a range outside the tree,
    /// so this guard is the difference between skipping one target and
    /// aborting the whole run.
````

</details>

File: crates/whisker-rust/src/provider.rs [hunk 12]
```diff
@@ -663,4 +808,26 @@ mod tests {
             "unexpected error: {error:#}"
         );
     }
+
+    /// A span past the 32-bit limit must produce no target at all
```
martijnarts, line 810: Rewritten. Rules broken:

- Rather-than ('confidently wrong rather than absent')
- -ing clauses 'Truncating', 'Reaching'
- Test doc wordier than the pin needs

````rust
    /// A span the 32-bit addressing cannot hold must produce no target at all
    ///
    /// Truncating instead would name a different, entirely unrelated span,
    /// and the decoration attached there would be confidently wrong rather
    /// than absent. Reaching the boundary through a real file would take a
    /// four-gigabyte fixture, which is why the arithmetic has its own seam.
````

</details>

File: crates/whisker-rust/src/decorations/fn_signature.rs [hunk 13]
```diff
@@ -1,57 +1,80 @@
-use crate::decorations::ResolvedType;
+use crate::decorations::{ErrorType, ResolvedType, ReturnMode};
 
 /// Function signature information
 ///
-/// Attached to function item nodes so lints can inspect the return type
-/// without re-querying the semantic model.
+/// Attached to `function_item` nodes so lints can inspect the return type
```
martijnarts, line 5: Rewritten. Rules broken:

- 33-word sentence
- Passive 'could be established', 'is attached'

````rust
/// Function signature information
///
/// Attached to `function_item` nodes so lints can inspect the return type
/// without re-querying the semantic model. A signature is attached to every
/// resolved function, even when nothing about it could be established: a
/// missing decoration reads identically to a function the provider never
/// reached, and those are different facts.
````

</details>

File: crates/whisker-rust/src/decorations/fn_signature.rs [hunk 14]
```diff
@@ -1,57 +1,80 @@
...
     }
 
-    /// Returns the resolved return type, if available
+    /// Returns the resolved return type, awaited for an `async fn`
     pub fn return_type(&self) -> Option<&ResolvedType> {
         self.return_type.as_ref()
     }
 
-    /// Returns the error type name for `Result`-returning functions
+    /// Returns the `E` of the `Result<T, E>` this function returns
```
martijnarts, line 35: Rewritten. Rules broken:

- 29-word sentence

````rust
    /// Returns the `E` of the `Result<T, E>` this function returns
    ///
    /// [`None`] means the function does not return a `core::result::Result`
    /// at all, including an `async fn` whose awaited type is not one, and
    /// one whose future could not be projected.
````

</details>

File: lints/anyhow_missing_context/src/lib.rs [hunk 15]
```diff
@@ -1,17 +1,27 @@
 use whisker_rust::RustLintPass;
-use whisker_rust::decorations::{FnSignature, ResolvedType};
+use whisker_rust::decorations::{FnSignature, ResolvedType, TypePathRef};
 use whisker_types::{DecoratedNode, Diagnostic, LintPass, RuleId, Severity};
 
 const RULE_ID: RuleId = RuleId("lint.anyhow-missing-context");
 
+/// The definition path of `anyhow::Error`
```
martijnarts, line 7: Rewritten. Rules broken:

- Summary is a riddle, does not name what the constant holds
- 25-word sentence with awkward 'can never match this however the type prints'

````rust
/// The one error type this rule is about
///
/// `anyhow` declares `Error` at its crate root, so its definition path and
/// its public path coincide and the module segment list is empty. A crate
/// with its own `Error` type, or one that does not depend on anyhow at all,
/// can never match this however the type prints.
````

</details>

File: lints/anyhow_missing_context/src/lib.rs [hunk 16]
```diff
@@ -1,17 +1,27 @@
...
+
 /// Flags uses of `?` on [`Result`] types without a preceding `.context()`
 /// or `.with_context()` call, but only when the enclosing function returns
 /// `Result<T, anyhow::Error>`
 ///
-/// Using `?` without `.context()` propagates errors without adding
-/// information about what operation failed. Rich error context makes
-/// debugging significantly easier by providing a chain of explanations
-/// for how the error occurred.
+/// A `?` without `.context()` propagates an error with no information
```
martijnarts, line 18: Rewritten. Rules broken:

- Significance inflation 'significantly easier'
- 32-word sentence in the added paragraph
- -ing opener 'Suggesting .context(..) there'

````rust
/// Flags uses of `?` on [`Result`] types without a preceding `.context()`
/// or `.with_context()` call, but only when the enclosing function returns
/// `Result<T, anyhow::Error>`
///
/// Using `?` without `.context()` propagates errors without adding
/// information about what operation failed. Rich error context makes
/// debugging significantly easier by providing a chain of explanations
/// for how the error occurred.
///
/// A `?` inside a closure, an `async` block, or a `try` block converts into
/// that body's error type, not the enclosing function's, so it is
/// attributed to the body and left alone. Suggesting `.context(..)` there
/// would propose a fix that does not compile.
///
/// [`Result`]: std::result::Result
````

</details>

File: lints/anyhow_missing_context/src/lib.rs [hunk 17]
```diff
@@ -70,11 +80,32 @@ impl RustLintPass for AnyhowMissingContext {
     }
 }
 
+/// Node kinds whose bodies own a `?`, so the ancestor walk stops at them
```
martijnarts, line 83: Rewritten. Rules broken:

- Em-dash aside '— and, worse, —'
- Rather-than ('rather than the enclosing function's')
- 33-word sentence

````rust
/// Node kinds that own a `?`, so the walk must not climb past them
///
/// A closure, an `async` block, a `gen` block, a `const` block, and a `try`
/// block each have their own return type, and `?` inside one converts into
/// that type rather than the enclosing function's. Attributing it to the
/// function would report an error type the expression never sees — and,
/// worse, would suggest `.context(..)` where it does not compile.
````

</details>

File: lints/anyhow_missing_context/src/lib.rs [hunk 18]
```diff
@@ -70,11 +80,32 @@ impl RustLintPass for AnyhowMissingContext {
...
+    "async_block",
+    "closure_expression",
+    "const_block",
+    "gen_block",
+    "try_block",
+];
+
 /// Walks up the tree to find the enclosing `function_item` and returns
 /// its [`FnSignature`] decoration, if present
+///
```
martijnarts, line 98: Rewritten. Rules broken:

- Em-dash aside pair around the false-negative clause
- -ing clauses 'on reaching ... before reaching'
- Hazy 'the right direction for a rule that runs in CI'

````rust
/// Walks up the tree to find the enclosing `function_item` and returns
/// its [`FnSignature`] decoration, if present
///
/// Returns [`None`] on reaching a body that owns the `?` before reaching a
/// function. That is a false negative by construction — a `?` inside a
/// closure that genuinely returns `anyhow::Result` is no longer flagged —
/// and it is the right direction for a rule that runs in CI.
````

</details>

File: lints/anyhow_missing_context/src/lib.rs [hunk 19]
```diff
@@ -190,6 +215,48 @@ mod tests {
...
+        map.insert(operand_id, result_type());
+        map.insert(fn_id, anyhow_fn_sig());
+        decorate(&mut tree, map);
+
+        let diagnostics = execute(&tree, &mut passes());
+
+        assert_no_diagnostics(&diagnostics);
+    }
+
+    /// A `?` inside a closure converts into the closure's error type
```
martijnarts, line 245: Rewritten. Rules broken:

- -ing opener 'Reporting', -ing clause 'closure returning something else'
- Near-duplicate of the compile rationale already in the rule doc

````rust
    /// A `?` inside a closure converts into the closure's error type
    ///
    /// Reporting the enclosing function's error type here would suggest
    /// `.context(..)` inside a closure returning something else, which does
    /// not compile.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 20]
```diff
@@ -145,6 +145,18 @@ fn flagged_within<'a>(
         .collect()
 }
 
+/// Returns the source text covered by every diagnostic in the file
```
martijnarts, line 150: Rewritten. Rules broken:

- 27-word convoluted sentence 'Listing the whole file is the expectation that has to be updated ...'
- -ing openers 'Listing', 'firing'
- 'which is what makes ... visible' construction

````rust
/// Returns the source text covered by every diagnostic in the file
///
/// A per-function expectation can only fail on a function it names, so it
/// cannot notice a rule that started firing somewhere else. Listing the whole
/// file is the expectation that has to be updated whenever the rule's reach
/// changes, which is what makes an unintended change in reach visible.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 21]
```diff
@@ -309,6 +321,37 @@ fn fn_signature_on_anyhow_result_function() {
     );
 }
 
+/// An `async fn`'s signature must describe the awaited type
```
martijnarts, line 326: Rewritten. Rules broken:

- Test doc over two sentences
- Figurative 'fail loudly'
- 29-word sentence

````rust
/// An `async fn`'s signature must describe what its body produces
///
/// `Function::ret_type` reads the signature output verbatim, which for an
/// `async fn` is the opaque future, not the type `?` in the body
/// converts into. Reverting to it leaves the return mode at `Direct` and the
/// error type at [`None`], so both halves of this expectation fail loudly.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 22]
```diff
@@ -613,6 +701,48 @@ fn anyhow_missing_context_flags_a_bare_try_in_an_anyhow_function() {
     );
 }
 
+/// An `async fn` that returns `anyhow::Result` is still an anyhow function
```
martijnarts, line 736: Rewritten. Rules broken:

- CLI-history aside is unrelated to what the test pins
- Idiom 'the rule gives up'
- -ing clauses 'returning', 'reading'

````rust
/// An `async fn` returning `anyhow::Result` is still an anyhow function
///
/// Its signature output is the opaque future, so a provider reading the
/// signature verbatim reports no error type and the rule gives up. Every
/// command in whisker's own CLI is `async`, which is why this shape is
/// pinned against the real provider instead of a hand-built signature.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 23]
```diff
@@ -613,6 +701,48 @@ fn anyhow_missing_context_flags_a_bare_try_in_an_anyhow_function() {
...
+
+    let diagnostics = whisker_core::walk(&tree, &mut passes);
+
+    assert_eq!(
+        flagged_within(&tree, &diagnostics, &func),
+        vec!["std::fs::read_to_string(\"async_anyhow_bare.txt\")?"]
+    );
+}
+
+/// An `async fn` in an impl block is reached the same way a free one is
```
martijnarts, line 757: Rewritten. Rules broken:

- -ing clause 'declares the method as returning'

````rust
/// An `async fn` in an impl block is reached the same way a free one is
///
/// The trait declares the method as returning `impl Future`, which is a
/// different shape from the `async fn` that implements it. The awaited type
/// has to come from the implementation, not the declaration.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 24]
```diff
@@ -630,6 +760,173 @@ fn anyhow_missing_context_flags_a_bare_try_on_a_method_call() {
     );
 }
 
+/// Only the anyhow function is flagged, out of four that all render `Error`
```
martijnarts, line 795: Rewritten. Rules broken:

- Unparseable essay, five sentences plus a second paragraph
- 30-word sentence 'This runs the real provider ... happen to agree with it'
- Figurative 'the sharp one'

````rust
/// Only the anyhow function is flagged, out of four that all render `Error`
///
/// `syn::Result`, `std::fmt::Result`, `std::io::Result`, and `anyhow::Result`
/// all put a type named `Error` in the `E` slot, and rust-analyzer renders
/// every one of them as `Error`. A rule deciding from the rendering flags all
/// four. This runs the real provider and the real rule over real source,
/// which is the only arrangement that can tell a fixed rule from a rule whose
/// hand-built fixtures happen to agree with it.
///
/// The `syn` case is the sharp one: `syn` defines `Error` in a private
/// `error` module and re-exports it at the root, so its definition path and
/// its public path differ.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 25]
```diff
@@ -630,6 +760,173 @@ fn anyhow_missing_context_flags_a_bare_try_on_a_method_call() {
...
+    assert_eq!(
+        flagged_within(&tree, &diagnostics, &anyhow_fn),
+        vec!["std::fs::read_to_string(\"anyhow_bare.txt\")?"]
+    );
+    assert!(flagged_within(&tree, &diagnostics, &syn_fn).is_empty());
+    assert!(flagged_within(&tree, &diagnostics, &fmt_fn).is_empty());
+    assert!(flagged_within(&tree, &diagnostics, &io_fn).is_empty());
+}
+
+/// An `async fn` that returns `std::io::Result` is not an anyhow function
```
martijnarts, line 827: Rewritten. Rules broken:

- Figurative 'trade one wrong answer for another'
- -ing opener 'Reading the awaited type', 33-word sentence

````rust
/// An `async fn` returning `std::io::Result` is not an anyhow function
///
/// Reading the awaited type has to preserve the discrimination, not just
/// restore the reach: a fix that made every `async fn` visible without also
/// identifying its error type would trade one wrong answer for another.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 26]
```diff
@@ -630,6 +760,173 @@ fn anyhow_missing_context_flags_a_bare_try_on_a_method_call() {
...
+    let func = find_function_by_name(&root, "returns_io_result_async")
+        .expect("should find returns_io_result_async");
+    let mut passes: Vec<Box<dyn LintPass>> = vec![AnyhowMissingContext::into_lint_pass()];
+
+    let diagnostics = whisker_core::walk(&tree, &mut passes);
+
+    assert!(flagged_within(&tree, &diagnostics, &func).is_empty());
+}
+
+/// A function that returns `std::io::Result` is outside this rule
```
martijnarts, line 845: Rewritten. Rules broken:

- Idiom 'not this rule's business'
- 'pins the identity question against the real provider' sentence duplicates the lookalike test's rationale
- -ing clauses 'returning', 'reading'

````rust
/// A `?` in a function returning `std::io::Result` is not this rule's business
///
/// `std::io::Result<()>` renders as `Result<(), Error>`, so a rule reading the
/// rendering sees a bare `Error` and cannot tell it from anyhow's. This pins
/// the identity question against the real provider, where the rendering is
/// produced, not hand-written.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 27]
```diff
@@ -630,6 +760,173 @@ fn anyhow_missing_context_flags_a_bare_try_on_a_method_call() {
...
+    let func =
+        find_function_by_name(&root, "returns_io_result").expect("should find returns_io_result");
+    let mut passes: Vec<Box<dyn LintPass>> = vec![AnyhowMissingContext::into_lint_pass()];
+
+    let diagnostics = whisker_core::walk(&tree, &mut passes);
+
+    assert!(flagged_within(&tree, &diagnostics, &func).is_empty());
+}
+
+/// `Box<dyn Error>` occupies the `E` slot as an ADT, and it is not anyhow's
```
martijnarts, line 863: Rewritten. Rules broken:

- Filler 'simply', idiom 'the one it cares about'

````rust
/// `Box<dyn Error>` occupies the `E` slot as an ADT, and it is not anyhow's
///
/// The `E` here is `Box`, whose definition lives in `alloc`, so the rule sees
/// a named error type that simply is not the one it cares about.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 28]
```diff
@@ -630,6 +760,173 @@ fn anyhow_missing_context_flags_a_bare_try_on_a_method_call() {
...
+    let func = find_function_by_name(&root, "returns_boxed_error")
+        .expect("should find returns_boxed_error");
+    let mut passes: Vec<Box<dyn LintPass>> = vec![AnyhowMissingContext::into_lint_pass()];
+
+    let diagnostics = whisker_core::walk(&tree, &mut passes);
+
+    assert!(flagged_within(&tree, &diagnostics, &func).is_empty());
+}
+
+/// An error type the signature leaves open cannot be anyhow's
```
martijnarts, line 881: Rewritten. Rules broken:

- 31-word sentence
- Filler 'may well', idiom 'stay quiet'

````rust
/// An error type the signature leaves open cannot be anyhow's
///
/// A caller may well instantiate `E` as `anyhow::Error`, but the body must
/// compile for every `E` the bounds admit, so `.context(..)` is not available
/// there and the rule must stay quiet.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 29]
```diff
@@ -630,6 +760,173 @@ fn anyhow_missing_context_flags_a_bare_try_on_a_method_call() {
...
+    let func = find_function_by_name(&root, "returns_generic_error")
+        .expect("should find returns_generic_error");
+    let mut passes: Vec<Box<dyn LintPass>> = vec![AnyhowMissingContext::into_lint_pass()];
+
+    let diagnostics = whisker_core::walk(&tree, &mut passes);
+
+    assert!(flagged_within(&tree, &diagnostics, &func).is_empty());
+}
+
+/// A crate's own `Error` type must not match anyhow's
```
martijnarts, line 900: Rewritten. Rules broken:

- Idiom 'answer to'
- 40-word sentence with trailing appositive
- History detail ('four of the seven') beyond what the pin needs

````rust
/// A crate's own `Error` must not answer to anyhow's
///
/// This is the shape that produced four of the seven false positives when
/// the rule was first run over whisker's own source: a crate declaring
/// `Error` at some path of its own, in a crate that need not depend on
/// anyhow at all.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 30]
```diff
@@ -630,6 +760,173 @@ fn anyhow_missing_context_flags_a_bare_try_on_a_method_call() {
...
+    let func = find_function_by_name(&root, "returns_local_error_result")
+        .expect("should find returns_local_error_result");
+    let mut passes: Vec<Box<dyn LintPass>> = vec![AnyhowMissingContext::into_lint_pass()];
+
+    let diagnostics = whisker_core::walk(&tree, &mut passes);
+
+    assert!(flagged_within(&tree, &diagnostics, &func).is_empty());
+}
+
+/// A `?` in a closure belongs to the closure, not to the function around it
```
martijnarts, line 918: Rewritten. Rules broken:

- -ing cluster 'removing the barrier fails it by gaining an entry instead of losing one'
- Last sentence unparseable

````rust
/// A `?` in a closure belongs to the closure, not to the function around it
///
/// The closure returns `std::io::Result`, so `.context(..)` on its `?` would
/// not compile; the function's own `read()?` is a genuine hit. The
/// expectation carries both, so removing the barrier fails it by gaining an
/// entry instead of losing one.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 31]
```diff
@@ -630,6 +760,173 @@ fn anyhow_missing_context_flags_a_bare_try_on_a_method_call() {
...
+    let func = find_function_by_name(&root, "closure_returning_io_result")
+        .expect("should find closure_returning_io_result");
+    let mut passes: Vec<Box<dyn LintPass>> = vec![AnyhowMissingContext::into_lint_pass()];
+
+    let diagnostics = whisker_core::walk(&tree, &mut passes);
+
+    assert_eq!(flagged_within(&tree, &diagnostics, &func), vec!["read()?"]);
+}
+
+/// Every place the rule fires in the fixture, listed once
```
martijnarts, line 937: Rewritten. Rules broken:

- Near-duplicates flagged_in_file's rationale sentence
- 33-word sentence
- Test doc over two sentences

````rust
/// Every place the rule fires in the fixture, listed once
///
/// A per-function expectation can only fail on the function it names, so the
/// suite stayed green while the binary reported `lint.anyhow-missing-context`
/// on a `std::io::Result` function nothing asserted about. One expectation
/// over the whole file puts the rule's reach under test, not its behavior
/// on four chosen functions.
````

</details>

File: crates/whisker-rust/tests/provider_integration.rs [hunk 32]
```diff
@@ -683,6 +980,23 @@ fn wildcard_match_arm_ignores_a_scrutinee_that_is_not_an_enum() {
     assert!(flagged_within(&tree, &diagnostics, &func).is_empty());
 }
 
+/// Counts the fixture's `function_item` nodes and how many carry a signature
```
martijnarts, line 1015: Rewritten. Rules broken:

- Em-dash aside pair '— the method call behind `?`, the `async` trait method —'
- 41-word sentence
- Inflation 'exactly the ones'

````rust
/// Counts the fixture's `function_item` nodes and how many carry a signature
///
/// The walk recurses because a method inside an `impl` block is a
/// `function_item` like any other, and the shapes this suite most depends on
/// — the method call behind `?`, the `async` trait method — are exactly the
/// ones a top-level-only loop never counts.
````

</details>

File: lints/anyhow_missing_context/src/lib.rs [hunk 33]
```diff
@@ -131,12 +159,14 @@ mod tests {
...
-        FnSignature::new(Some(ret), Some("anyhow::Error".into()))
+        let error = ErrorType::Named(TypePath::new("anyhow", [] as [&str; 0], "Error"));
+        FnSignature::new(Some(ret), Some(error), ReturnMode::Direct)
     }
 
     fn non_anyhow_fn_sig() -> FnSignature {
         let ret = ResolvedType::new("Result<(), std::io::Error>".into()).with_result(true);
-        FnSignature::new(Some(ret), Some("std::io::Error".into()))
+        let error = ErrorType::Named(TypePath::new("core", ["io", "error"], "Error"));
+        FnSignature::new(Some(ret), Some(error), ReturnMode::Direct)
```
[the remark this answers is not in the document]
martijnarts: You found a real inconsistency and blamed the wrong half of it. The crate
name is right. On this toolchain `std::io::Error` is a re-export: `core` declares the
struct in `library/core/src/io/error.rs`, so the definition path is
`core::io::error::Error`. I did not reason that out, I measured it.
`fn_signature_on_io_result_function` now asserts the path that
`returns_io_result` produces, and it passes with `core`.

The rendering beside it was the part that was wrong. The provider renders both
`anyhow::Result<()>` and `std::io::Result<()>` as `Result<(), Error>`, never
with a path, so `"Result<(), std::io::Error>"` was a string rust-analyzer
never produces. Both fixtures now use `Result<(), Error>`, which also shows
what the test is for: the two signatures differ only in the definition path.

You refer to "the comment", but the fixture had none. The display string was
the only thing claiming `std`. Each fixture now carries a doc comment that
names the signature it builds.

PR #220, opened 2026-08-05
Route per-file check failures through one mode

martijnarts:
The check loop had two places that decided what to do with a file it could not
process: one for a read that fails, one for a pipeline run that fails. Each
re-derived the meaning of `keep_going` from the raw flag. Only the read site
was pinned by a test, so deleting the failure recording from the analysis site
left the suite green, and nothing stopped the two branches from drifting into
disagreement about whether a failure reaches the exit code.

Both sites now call `ErrorRecovery::record`. One place turns the mode into
behavior, and the flag becomes that mode where the arguments are destructured
instead of being re-interpreted deep in the loop. The aborting path also gained
the file name as error context, which it was missing whenever the pipeline was
the thing that failed.

The old keep-going test asserted only on the underlying I/O message, which both
paths print and both exit 1 on, so it passed even with the flag's whole
recovery branch removed. It now names the line only the recovering path emits,
and a companion test pins the aborting path it has to be told apart from.

### DIFF
--- a/crates/whisker/src/commands/check/error_recovery.rs
+++ b/crates/whisker/src/commands/check/error_recovery.rs
@@ -1,0 +1,116 @@
+use std::path::Path;
+
+use super::check_outcome::CheckOutcome;
+
+/// What `whisker check` does with a file it cannot read or analyze
+///
+/// Read failures and analysis failures both go through
+/// [`ErrorRecovery::record`], so one place decides whether a per-file
+/// failure ends the run. The `--keep-going` flag selects
+/// [`ErrorRecovery::Continue`].
+// r[impl cli.check.keep-going]
+#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
+pub(crate) enum ErrorRecovery {
+    /// The first failing file ends the run with its error
+    Abort,
+    /// Whisker reports each failing file, continues, and still fails the run
+    Continue,
+}
+
+impl ErrorRecovery {
+    /// Records a failure for one file
+    ///
+    /// Under [`ErrorRecovery::Continue`], the error goes to stderr and
+    /// `outcome` becomes [`CheckOutcome::Failure`].
+    ///
+    /// # Errors
+    ///
+    /// Under [`ErrorRecovery::Abort`], returns `error` with the file name
+    /// added, so the caller can end the run with it.
+    pub(crate) fn record(
+        self,
+        outcome: &mut CheckOutcome,
+        file: &Path,
+        error: anyhow::Error,
+    ) -> anyhow::Result<()> {
+        match self {
+            Self::Abort => Err(error.context(format!("check {}", file.display()))),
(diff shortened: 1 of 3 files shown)


martijnarts:
Every doc comment this PR adds was re-checked against three rule sets and rewritten where it failed. 3 blocks changed here.

Violations found here: 2 ASD-STE100, 1 AI-writing tells.

|---|---|---|
| `crates/whisker/src/commands/check/error_recovery.rs` | ErrorRecovery::Abort variant | Passive voice (its error is returned) |
| `crates/whisker/src/commands/check/error_recovery.rs` | ErrorRecovery::Continue variant | Passive voice (a failing file is reported) |
| `crates/whisker/src/commands/check/error_recovery.rs` | enum ErrorRecovery | Figurative language (reads and pipeline runs 'fail through' this type) |

crates/whisker/src/commands/check/error_recovery.rs
```diff
@@ -0,0 +1,116 @@
+use std::path::Path;
+
+use super::check_outcome::CheckOutcome;
+
+/// What `whisker check` does with a file it cannot read or analyze
```
martijnarts, line 5: Rewritten. Rules broken:

- Figurative language (reads and pipeline runs 'fail through' this type)

````rust
/// What `whisker check` does with a file it cannot read or analyze
///
/// Reads and pipeline runs both fail through this type, so one place
/// decides whether a per-file failure ends the run. The `--keep-going`
/// flag selects [`ErrorRecovery::Continue`].
````

</details>

File: crates/whisker/src/commands/check/error_recovery.rs [hunk 2]
```diff
@@ -0,0 +1,116 @@
...
+///
+/// Read failures and analysis failures both go through
+/// [`ErrorRecovery::record`], so one place decides whether a per-file
+/// failure ends the run. The `--keep-going` flag selects
+/// [`ErrorRecovery::Continue`].
+// r[impl cli.check.keep-going]
+#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
+pub(crate) enum ErrorRecovery {
+    /// The first failing file ends the run with its error
```
martijnarts, line 14: Rewritten. Rules broken:

- Passive voice (its error is returned)

````rust
    /// The first failing file ends the run and its error is returned
````

</details>

File: crates/whisker/src/commands/check/error_recovery.rs [hunk 3]
```diff
@@ -0,0 +1,116 @@
...
+    Abort,
+    /// Whisker reports each failing file, continues, and still fails the run
```
martijnarts, line 16: Rewritten. Rules broken:

- Passive voice (a failing file is reported)

````rust
    /// A failing file is reported, the run continues, and the run still fails
````

</details>

martijnarts (commented):

PR #238, opened 2026-08-11
Cut the test job from eight minutes to under three

martijnarts:
The `test-rust` job took 7m50s. Only 87s of that was compilation; the rest was two integration suites, `check.rs` at 258s for 23 tests and `provider_integration.rs` at 96s for 40. Both spend their time inside rust-analyzer, and a one-file fixture costs about as much as a real crate, because the cost is loading a project rather than analyzing it.

Three changes, in order of what they buy:

**`opt-level = 2` for dependencies.** An unoptimized `ra_ap_*` runs slowly, and the suite runs it far more often than it compiles it. The 23 CLI tests went from 85.6s to 36.4s, for 26 extra seconds of compilation.

**Nextest instead of `cargo test`.** The slow suites overlap rather than running one binary after another. This also makes `.config/nextest.toml` and `NEXTEST_PROFILE` take effect — the workflow sets that variable today and nothing reads it. Nextest cannot run doctests, so those follow in a second command rather than quietly disappearing; all 13 doctest sections still run.

**Caching.** Every job currently rebuilds all 265 dependency crates from scratch. The key follows `Cargo.lock` and `rust-toolchain.toml` rather than `rustc`, because the runner preinstalls a stable toolchain that no build here uses, so a compiler-derived key would miss the nightly bumps that land most days.

Locally the full 546-test suite now runs in 56s, bounded by its slowest single test at 47s. CI on this PR is the real measurement.

One caveat worth watching: the cache is keyed on `rust-toolchain.toml`, which renovate bumps daily, so the cache is invalidated about as often as it is used. Slowing that cadence would compound this change, and nothing in the tree needs a current nightly — no `#![feature]`, no `rustc_private`, only one unstable rustfmt option.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,4 +1,14 @@
 {
   $schema: "https://docs.renovatebot.com/renovate-schema.json",
   extends: ["github>aonyx-ai/renovate-config"],
+  packageRules: [
+    {
+      // Nothing here needs a current nightly: no `#![feature]`, no
+      // `rustc_private`. The pin exists for one unstable rustfmt option, so a
+      // daily bump only invalidates the CI cache.
+      matchPackageNames: ["rust"],
+      matchDepTypes: ["toolchain"],
+      schedule: ["on the first day of the month"],
+    },
+  ],
 }
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -3,9 +3,6 @@
   extends: ["github>aonyx-ai/renovate-config"],
   packageRules: [
     {
-      // Nothing here needs a current nightly: no `#![feature]`, no
-      // `rustc_private`. The pin exists for one unstable rustfmt option, so a
-      // daily bump only invalidates the CI cache.
       matchPackageNames: ["rust"],
       matchDepTypes: ["toolchain"],
       schedule: ["on the first day of the month"],
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,23 @@ jobs:
       - name: Remove Flox installer (if it exists)
         run: rm -f flox.x86_64-linux.deb
 
+      # Every job starts from an empty runner, so without this each one
+      # rebuilds all 265 dependency crates. The key follows `Cargo.lock` and
+      # `rust-toolchain.toml`, which together decide what gets built. The
+      # runner preinstalls a stable toolchain that no build here uses, so a
+      # key derived from `rustc` instead would miss the nightly bumps that
+      # land most days.
+      - name: Cache Cargo downloads and build output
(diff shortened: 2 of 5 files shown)


.github/renovate.json5
```diff
@@ -1,4 +1,14 @@
 {
   $schema: "https://docs.renovatebot.com/renovate-schema.json",
   extends: ["github>aonyx-ai/renovate-config"],
+  packageRules: [
+    {
+      // Nothing here needs a current nightly: no `#![feature]`, no
+      // `rustc_private`. The pin exists for one unstable rustfmt option, so a
+      // daily bump only invalidates the CI cache.
```
martijnarts: this comment is useless and annoying

.github/workflows/ci.yml
```diff
@@ -66,12 +66,8 @@ jobs:
...
         run: rm -f flox.x86_64-linux.deb
 
-      # Every job starts from an empty runner, so without this each one
-      # rebuilds all 265 dependency crates. The key follows `Cargo.lock` and
-      # `rust-toolchain.toml`, which together decide what gets built. The
-      # runner preinstalls a stable toolchain that no build here uses, so a
-      # key derived from `rustc` instead would miss the nightly bumps that
-      # land most days.
+      # Keyed on the lock file and the toolchain, not on `rustc`: the runner
+      # preinstalls a stable toolchain that no build here uses.
```
martijnarts: you're doing "Y, not X" without explaining X, when you shouldn't even be talking about X.

Cargo.toml
```diff
@@ -27,11 +27,8 @@ ignore = [
 missing_errors_doc = "warn"
 missing_panics_doc = "warn"
 
-# Build dependencies with optimizations even in debug builds. The test suite
-# spends most of its time inside rust-analyzer rather than inside whisker, so
-# an unoptimized `ra_ap_*` dominates every run that loads a project. Measured
-# on the 23 CLI tests: 85.6s unoptimized against 36.4s here, for 26 extra
-# seconds of compilation.
+# The tests run rust-analyzer far more than they compile it. Optimizing
+# dependencies took the CLI suite from 85.6s to 36.4s.
```
martijnarts: specific timing is not context for code comments

justfile
```diff
@@ -68,9 +68,7 @@ prettier fix="false" extension="*":
 
 # Run the tests
 #
-# Nextest gives each test its own process, so the slow integration suites
-# overlap instead of running one binary after another. It cannot run doctests,
-# so those follow separately.
+# Nextest cannot run doctests, so those follow separately.
```
martijnarts: this doesn't work with just. remove it.

.github/renovate.json5
martijnarts: Removed.

.github/workflows/ci.yml
martijnarts: Removed.

PR #277, opened 2026-08-22
Move the rules to a repository of their own

martijnarts:
The rules move out of this repository and into
[aonyx-ai/whisker-aonyx-rules][rules]. Whisker also gets the function that
makes the move possible. A `[[lints]]` entry can now name a repository and a
commit. Before this change, an entry could name only a directory on the local
disk.

Whisker accepts a full commit hash, and it accepts nothing else. It does not
accept a branch or a tag. A branch can point to a different commit on a
different day. The same configuration would then run different rules and give
different results. A tool that changes its results without an instruction is
worse than a tool that is out of date.

A commit hash always names the same files, and whisker uses this fact. If a
checkout is already on the disk, whisker keeps it and does not examine it
again. Thus only the first check with a given pin uses the network.

Whisker gets the commit with the gitoxide library. It does not run the `git`
program. It asks the remote for that one commit at a depth of one. A check
thus does not depend on the version of git on the machine. A check also does
not fail if the machine has no git program.

A configured directory can now contain a cargo workspace. Whisker loads each
dynamic library that the build makes. Whisker then does the ABI handshake for
each library. One entry is thus enough for a repository that contains 19
rules. This applies also to a path entry, which the related code refused
before this change.

The move of the rules made changes to the tests necessary. The whisker-rust
crate had three of the rules as dev-dependencies. Its provider tests used
those rules to examine diagnostics instead of decorations. The tests in
`check.rs` used a fourth rule when a fixture needed a rule that gives a
warning.

These tests measured the output of the provider. They showed that a real
`FnSignature` gives the error type that a rule compares. They also showed that
`ImportSource` can tell the difference between `Color::Red` and `Shapes::draw`.
The new `decoration_probes` package contains the decoration code of each of
these rules. It does not contain the conditions that make each rule give a
warning. Each test gives the same result as before.

Three tests in `check.rs` had a different problem. Their fixtures have no
configuration file. Whisker looks in each parent directory for a configuration
file until it finds a `.git` directory. Thus these three tests read the
configuration of this repository and ran the rules that it named.

These three tests thus tested less than they appeared to test. With the rules
in a remote repository, they would also use the network during `cargo test`.
Each fixture now names its own rules. The recipe `just check-self` replaces
the test that checked the current directory.

Each commit builds and passes the tests on its own. The first three commits
are the git source function, the test changes, and a format pass on the plugin
packages. The workspace boundary had kept those packages away from `cargo
fmt`. The fourth commit removes the rules and changes the configuration. The
last commit corrects two problems that CI found.

A local machine cannot find the first of these problems. The code detached
`HEAD` with a ref transaction. A ref transaction writes a reflog, and a reflog
needs a committer. Thus the fetch operated correctly on each machine that has
a `~/.gitconfig` file, but failed on each build agent. The tests now run with
an empty git configuration to prevent the same problem in the future.

The second problem is the license policy. The `deny.toml` file gets three more
licenses, and this is the one policy decision in this change. BSD-3-Clause and
MIT-0 come from the TLS code of gitoxide and from the text filters of the
checkout. CDLA-Permissive-2.0 applies to the list of certificate authorities
from Mozilla, not to code.

No copyleft license is in the list. The pack caches of gitoxide would have
added MPL-2.0, but those caches do not help a single shallow fetch. Whisker
thus names only the one feature that it needs from that group.

One task stays for later. The rules repository points to this branch now,
because `main` does not have the git source function yet. After this branch
merges, that pin must move to the merge commit. The `.config/whisker.toml`
file in this repository must move with it.

[rules]: https://github.com/aonyx-ai/whisker-aonyx-rules


martijnarts:
A stack of 13 smaller pull requests replaces this one: #279 through #292.
The end state is the same, with four differences.

Three of them answer the Copilot review on this pull request. `GitUrl` now
removes the userinfo from a remote before Display or Debug prints it, and
the two configuration errors that name a remote go through that same path.
The wildcard probe now tests both answers of the `AdtFlags` decoration. The
documentation of `holds` no longer claims to detect a manual edit, because
it does not; it now says what the atomic install does guarantee, and where
the trust boundary is.

The fourth difference is two new tests. One drives `whisker check` against
a path entry that names a cargo workspace, and shows that two members with
different rules both report. The other shows what a git entry says when the
remote cannot be reached.

Please do not delete the `git-sourced-lints` branch yet. The rules
repository pins whisker at 52037ae, which is a commit on that branch. After
the stack merges, that pin must move to a commit on main.

PR #279, opened 2026-08-23
Let taplo settle the workspace exclusions

martijnarts:
The exclude array in the workspace manifest was on one long line. taplo
rewrites that array each time a recipe formats TOML. Every commit thus
carried an unrelated manifest change, and the pre-commit hook stopped the
commit. This change applies the format one time. The array now has the
shape that the formatter wants.

This change contains no code.


martijnarts:
@comment-quality review

PR #305, opened 2026-08-27
Publish release archives for three targets

martijnarts:
Whisker ships no binaries today. A person who wants to run it must build it first. That build needs the pinned nightly toolchain and several minutes.

This change attaches a compressed archive to every GitHub release. The archives cover Linux on x86-64, Linux on arm64, and macOS on Apple silicon. Each archive holds the binary, both licenses, and the README. A sidecar file beside each archive holds a SHA-256 digest, so a user can check a download with `shasum -a 256 -c`.

The new workflow replaces the workflow that published to crates.io. That workflow called `just publish`, and this repository never had that recipe. Every crate here also sets `publish = false`. A crates.io job can return when there is something to publish.

The recipe that assembles an archive rejects a tag that disagrees with `Cargo.toml`. Such a tag produces an archive whose name promises a version that the binary does not report.

The job keeps no Cargo cache, although every other workflow here restores one. A pull request can write to a cache, and that cache then reaches the binary that a release publishes. A release builds once per tag, so the saved minutes are worth less than that guarantee. The job also uses rustup rather than Flox, because the Flox action installs on x86-64 Linux alone.

---


martijnarts:
@comment-quality review

martijnarts:
@comment-quality review

PR #306, opened 2026-08-27
Move the lint cache beside the sources that fill it

martijnarts:
The cache module lived inside the git source module, because the git source was the only thing that fetched anything. Whisker gains a second way to obtain a lint source later in this stack. That second way needs the same cache root, the same directory per remote, and the same install by rename.

This change lifts the module one level. The cache now sits beside the sources rather than inside one of them. Nothing else moves. The layout, the environment override, and every test stay the same.

The module also gains documentation of its own. It had none while it was a private detail of a single caller.

---


martijnarts:
@comment-quality review

martijnarts:
@comment-quality review

PR #310, opened 2026-08-27
Document installation and prebuilt lints

martijnarts:
Whisker now ships binaries, and it loads lints that a publisher built. Neither fact appears anywhere a user reads.

The README gains an installation section, and the custom lints section gains the prebuilt path. It states what whisker asks a release for, what it does with the answer, which variables reach a private repository, and what the digest beside an archive proves. It also states that a released whisker accepts only a library built by the nightly that built it, and what to do about that.

ARCHITECTURE derives the tag from what the handshake compares, so a whisker nobody built for finds no file. It also explains why a prebuilt library that fails the handshake ends the run, and which failures print a warning.

---

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -119,6 +119,60 @@ revalidated against the remote, which is what keeps the network out of the
 common path. Only a git entry builds with `--locked`: the pin is worth
 what the lockfile behind it is worth.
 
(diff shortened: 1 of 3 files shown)


martijnarts:
@comment-quality review

martijnarts:
@comment-quality review

PR #332, opened 2026-09-03
Let a project run a subset of a source's rules

### LINKED ISSUE #331: Let a project run a subset of a source's rules
A project should be able to run a subset of the rules a lint source provides.

Today a `[[lints]]` entry loads every plugin the source builds, and the only other lever is `--deny-warnings`, which applies to all of them. A project therefore takes all of a source's rules or none.

That is what stops a project adopting whisker gradually. raft is the first outside repository pointed at the Aonyx rules, and it reports 2,995 warnings across fifteen rules. Its choices are to fix all of them before the check can block, or to run a check that cannot fail. Neither is a good first step, and the second is worse: a green check that reports thousands of warnings is read by nobody.

There is a path today, and it is deliberate rather than a workaround. Every rule crate sets `default = ["plugin"]`, and turning that feature off makes the rule a plain rlib. A project can therefore keep a crate of its own that depends on the rules it wants and calls `export_lints!` with them. It costs a crate per project, and it forfeits the prebuilt archives, because the selection is not what any publisher built.

The work:

- A configuration surface that names rules to enable or disable, by rule id.
- Applied after the handshake, so a disabled rule is one whose diagnostics are dropped rather than one whose library is not loaded. A source still builds and loads as one unit.
- A decision on the default: every rule a source provides, or none until named.
- An error for a name that matches no rule the loaded sources provide, because a typo that silently disables nothing looks the same as a rule that never fires.

Per-rule severity is the neighbouring feature and can be considered with it. A project that wants a rule visible without blocking is asking for the same thing from the other side.

martijnarts:
Closes #331.

A `[[lints]]` entry brings every rule its source provides, and the only other lever was `--deny-warnings`. A project took all of a source's rules or none. That is what stopped raft adopting whisker: of the 2,995 warnings there, 365 come from one rule flagging the `s[impl ...]` markers its own spec-coverage gate reads out of comments, so fixing them would break `just check-specs`.

A `[rules]` table now names what runs:

```toml
[rules]
disable = ["lint.no-inline-comments"]
```

`enable` is the other half and runs only the rules named, which is how a project takes on one rule at a time. Naming both is refused. A name that no loaded plugin declares fails the run, because a misspelling would otherwise disable nothing and say nothing, which reads exactly like a rule that found no fault.

That check needs plugins to declare their rules, and where that goes decides what the upgrade costs.

My first attempt put a `rules` method on `LintPass`. That reorders a vtable, which whisker cannot measure, so every plugin ever built would have been refused and every published archive stranded. Appending a field to `PluginDeclaration` is different: the struct is `#[repr(C)]`, whisker already reads `abi_version` at offset 0 through a raw pointer before forming a reference, and so it knows where each version's fields end. A plugin that stops sooner offers less rather than being turned away.

So the protocol is 3 and whisker reads 2 upward. The consequences, all tested:

- A plugin exporting the protocol-2 declaration loads, and its rules run. There is a fixture that spells that older struct out by hand, because it is the shape a real pre-3 plugin exports.
- Such a plugin declares no rules, so naming one in `[rules]` fails rather than being silently accepted.
- The tag an archive is published under now carries `MIN_ABI_VERSION` rather than `ABI_VERSION`, so two whiskers that read the same protocols accept each other's archives.

The last one is what makes this cheap rather than expensive: **the archives already published for `v0.1.0-rc.1` are still found and still load.** I checked against the live release rather than a fixture — a cold cache downloads `f9fc0bc-2d20b32c659070af-aarch64-apple-darwin.tar.gz`, unpacks it, loads nineteen protocol-2 libraries, and reports. `check-self` passes unchanged, so none of the cross-repo sequence I described earlier is needed.

Follow-up this does not do: per-rule severity, which is the same request from the other side.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -283,6 +283,21 @@ declaration's `ABI_VERSION`. A test scans both traits' source and fails
 when either method list moves, so the test reminds a contributor to bump
 it.
 
+Two kinds of change reach that declaration, and they cost differently. A
+field appended to it is readable by version, because the struct is
+`#[repr(C)]` and whisker knows where each version's fields end. A plugin
+that stops sooner offers less rather than being refused. A method added
+to either trait is not readable that way, because whisker cannot measure
+a vtable. So an optional capability belongs in the declaration, and
+`MIN_ABI_VERSION` names the oldest layout whisker still reads. Raising
+`ABI_VERSION` alone keeps older plugins loading; raising the floor is what
+stops them, and only a vtable change needs that.
+
+The tag a publisher names an archive with carries the floor rather than
+the version, for the same reason. Two whiskers that read the same
+protocols accept each other's archives, so a protocol that only appends a
+field does not strand what a publisher already built.
+
 The fingerprints stop at whisker's own layout. A contributor maintains
 each list of types by hand. A fieldless enum keeps its size and alignment
 when its variants reorder, so the fingerprint does not see that change. A
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to
 
 ### Added
 
+- A `[rules]` table names the rules a project runs. `enable` runs only those
+  named, `disable` runs everything else, and a name that no configured lint
+  reports is an error rather than a filter that quietly admits everything.
 - Whisker asks a git source's releases for prebuilt lints before it compiles
   anything. It verifies the published SHA-256, and every library still
   completes the plugin handshake.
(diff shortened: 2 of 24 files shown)


martijnarts:
Copilot is right, and this was the bug that mattered most in the PR.

`load_library` formed a `&PluginDeclaration` as soon as the version was one whisker accepts. That was sound while whisker accepted exactly one version. It stopped being sound in this PR: a protocol-2 plugin exports a static that ends after `register`, so the reference claims eight bytes that are not there. It also claims every field holds a valid value, and `rules` is a function pointer the compiler may assume is not null. Both claims are false whether or not the field is read, so guarding the read of `rules` was not enough.

The loader now reads each field through a raw pointer at its own offset and projects a field only once the version says the plugin exported it. No step depends on the plugin agreeing about the struct's size.

This is worth spelling out because the doc comment on `load_library` already described the correct rule and the code did not follow it. Tests could not catch it: reading eight bytes past a static lands in the same section, so nothing faults, and Miri cannot reach this path because it involves `dlopen` of a real cdylib. The new test pins the invariant that is checkable instead, which is that `rules` sits after every field protocol 2 exported. A field inserted among them would shift the rest and whisker would read one field's bytes as another, silently.

Fixed in 4e6bd0b.

